PR #6, opened 2018-05-31
Added spec file for CentOS 7 & systemd units

sklochkov:
Added a spec file for CentOS 7 build & normal and instanced units for systemd.


pmwkaa:
Thank you! Can we move those files to scripts/centos/ directory or they must be put in root directory to work?

sklochkov:
Since systemd is used not only on CentOS, it would make sense to put the unit files to a systemd/ directory. I'll need to update the spec file for that to work.

Also, what's the current version of odyssey? I found 3.0 in the debian/ directory.

sklochkov:
The spec file can be put anywhere, it will work from anywhere if the correct tarball is present in rpmbuild/SOURCES

pmwkaa:
scripts/systemd/ sounds good to me, lets use it. 

I believe this should be 1.0 beta. I think to stick to semantic versioning (https://semver.org/) should be a good idea

sklochkov:
Ok, then it would make sense to change the Version line in the spec file.

sklochkov:
Done.

PR #28, opened 2018-06-12
Fix show stats database string format

ikitiki:
There's a null-terminator in the database field of the "show stats" output


pmwkaa:
I wonder how i missed that, thanks!

PR #43, opened 2018-10-08
Add pause/resume admin console commands

tomlikestorock:
Adds `PAUSE`/`RESUME` admin console commands.

This puts pause/resume `state` on `config_storage`s. Any attempt to use a route with a paused storage is blocked for 1 second then retried, and any idle servers with paused storage are marked for gc.

`PAUSE` can be called with or without the name of a storage to pause. If no storage name is given, all storages will be paused. `RESUME` operates similarly.

Open to comments/critiques; wasn't 100% confident on how to block attempts to attach. Initially tried to aim for a way to let clients still be able to connect but not be given a server until `RESUME`, but that seemed to create something too fragile.

### DIFF
--- a/sources/config.c
+++ b/sources/config.c
@@ -778,6 +778,7 @@ od_config_validate(od_config_t *config, od_logger_t *logger)
 		}
 		if (strcmp(storage->type, "remote") == 0) {
 			storage->storage_type = OD_STORAGE_TYPE_REMOTE;
+			storage->state = OD_STORAGE_ACTIVE;
 		} else
 		if (strcmp(storage->type, "local") == 0) {
 			storage->storage_type = OD_STORAGE_TYPE_LOCAL;
--- a/sources/config.h
+++ b/sources/config.h
@@ -44,11 +44,18 @@ typedef enum
 	OD_STORAGE_TYPE_LOCAL
 } od_storage_type_t;
 
+typedef enum
+{
+	OD_STORAGE_ACTIVE,
+	OD_STORAGE_PAUSE
+} od_storage_state_t;
+
 struct od_config_storage
 {
 	char              *name;
 	char              *type;
 	od_storage_type_t  storage_type;
+	od_storage_state_t state;
 	char              *host;
 	int                port;
 	od_tls_t           tls_mode;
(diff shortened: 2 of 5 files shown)


tomlikestorock:
This is good feedback, thank you. 

1. From what I gathered in the pgbouncer docs any currently established sessions or transactions are allowed to continue to exist until they naturally end via `COMMIT` or closing of the session, and this causes the `PAUSE` command to block until they are done. I attempted to recreate this behavior because I wanted a graceful end to existing connections, and not kill them in the middle of whatever they were doing. I didn't look into prepared statement management at all, pgbouncer seems to not make it easy there either so I figured that could be refined later.

2. I'll look into this

3. Oh, this blocks the admin console for all those connected to it? I wasn't quite sure how sleep in machinarium worked, and I could only find a usage of `machine_condition` used, which I wasn't confident in what it did differently. Could you explain the difference? Either way, the idea of pushing the console action into a coroutine makes sense to me, I'll look into it.

4. Originally I put `state` on the routes used, which turned out to be an obvious mistake in hidesight, and the config seemed more persistent. However, I forgot about config reloading. I'll look into moving it to the server object, or something else.

I'll keep adding commits to this PR if you're okay with doing the code review for it!

pmwkaa:
Yes, sure. I am ready to review more updates or provide any additional help.

1. I think we should just explicitly support only transactional pooling here, otherwise this will not work for clients who wants session pool guarantees: http://pgbouncer.github.io/features.html (SQL feature map for pooling modes).

3. `machine_sleep()` will switch current coroutine execution context to any other one, which is ready. Console subsystem is driven by single coroutine here: https://github.com/yandex/odyssey/blob/master/sources/console.c#L869. It simply loops https://github.com/yandex/odyssey/blob/master/sources/console.c#L829 and waits for new console requests. We probably could create a new coroutine here: https://github.com/yandex/odyssey/blob/master/sources/console.c#L839 instead of executing the request directly. Coroutine will exit automatically after return from its main function. We should be aware that there are possible situation when several coroutines executing console queries and accessing shared system data. This should be not a problem, pretty much all other subsystems (cron, system, router) are doing the same.

4. Probably it is also possible to derive config state, when new version is created.

x4m:
@tomlikestorock thanks for starting this work. I'm closing the issue since we merged #66.

PR #46, opened 2018-12-22
Implement SHOW POOLS command

x4m:
Hi!
Here's a stub for SHOW POOLS command at cosole.
It's a WiP, there are few important things like ```maxwait``` for which I did not figured out how to easily get it.
Thanks!

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -4,3 +4,4 @@ published and distributed by YANDEX LLC as the owner:
 Dmitry Simonenko <pmwkaa@yandex-team.ru>
 Dmitry Sarafannikov <dsarafan@yandex-team.ru>
 Evgeniy Efimkin <efimkin@yandex-team.ru>
+Andrey Borodin <x4mmm@yandex-team.ru>
--- a/sources/console.c
+++ b/sources/console.c
@@ -26,7 +26,8 @@ enum
 	OD_LSERVERS,
 	OD_LCLIENTS,
 	OD_LLISTS,
-	OD_LSET
+	OD_LSET,
+	OD_LPOOLS
 };
 
 static od_keyword_t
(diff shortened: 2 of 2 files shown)


pmwkaa:
Hi, thanks for contribution. Generally this looks good, except few things:

1. Route pool and route object information must not be accessed directly here: https://github.com/yandex/odyssey/pull/46/commits/44647e8c4155934741868d13c6a4c32ccb9e0572#diff-9313c985cfb35d45dc2da8f7d957cb74R367 and here https://github.com/yandex/odyssey/pull/46/commits/44647e8c4155934741868d13c6a4c32ccb9e0572#diff-9313c985cfb35d45dc2da8f7d957cb74R265. Otherwise this could lead to crush when parallel client tries to update the data. Everything should be done via router, like here: https://github.com/yandex/odyssey/blob/44647e8c4155934741868d13c6a4c32ccb9e0572/sources/console.c#L719 and here https://github.com/yandex/odyssey/blob/44647e8c4155934741868d13c6a4c32ccb9e0572/sources/console.c#L669. This should be easy to fix.

2. `maxwait` according to PgBouncer sources is `how long is the oldest client waited`. Basically this show maximum query wait time of a oldest client in pool wait queue. It is a bit tricky, this would only work when pool_size > 0 and only for multi-statement transactions now only (because query_time_start calculated after request being sent to a server). I suggest to implement it as a separate patch.

3. few minors: missed `else` here? https://github.com/yandex/odyssey/pull/46/commits/44647e8c4155934741868d13c6a4c32ccb9e0572#diff-9313c985cfb35d45dc2da8f7d957cb74R324. We start labels from beginning of line: https://github.com/yandex/odyssey/pull/46/commits/44647e8c4155934741868d13c6a4c32ccb9e0572#diff-9313c985cfb35d45dc2da8f7d957cb74R333.

PR #55, opened 2019-04-03
Handle skipped results correctly during connection deploy

x4m:
Connection reconfiguration during deploy may issue queries. These queries need to be read from the server, but not shown to the client.
These queries were not handled correctly, causing a connection leak.

Also, it seems like that client HW read errors will lead to similar behaviour.


x4m:
This was fixed by recent commit, closing.

PR #57, opened 2019-04-25
Use memory contexts for allocated objects

ildus:
This pull request adds support of memory contexts for allocated objects. This is first revision of it, and I have moved to memory contexts only od_ objects for now, but it will have meaning for `kiwi` and `machinarium` objects too.

The idea is simple, when we have some scope (client, instance, config, route) it will have associated memory context with it, and all related objects will be allocated in it (including base object). And when we clean for example some `route`, we just remove its memory context and clean all allocated objects in this memory context.

I will be glad to hear some opinions about this pull request.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,5 @@
-cmake_minimum_required(VERSION 2.8)
-
-project(odyssey C)
+cmake_minimum_required(VERSION 3.5)
+project(odyssey LANGUAGES C)
 
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH}
     ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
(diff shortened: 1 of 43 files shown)


pmwkaa:
Hey, looks like a big pile of work! Thanks for the interest in the project.

Currently Odyssey rely on Machinarium for low-level management which intentionally should include memory management as well. From Machinarium perspective allocated objects (messages for example) are cached and reused after being freed to reduce interactions with system allocator. Since work is done in multi-threaded environment Machinarium keeps message cache associated with each created thread separately to reduce locking and control memory pools usage of malloc. Basically Odyssey tries to avoid system memory allocations and rely on cache when possible.

Machinarium interaction is based around sending and receiving messages. This might be an issue with introduced Postgres style pool allocations. So I am saying, that this approach might not cover all cases. But still can be very convenient in case if we want to introduce some kind of functionality which includes DBMS style processing to Odyssey.

Do you have such plans? :smiley_cat:

ildus:
Basicaly I was thinking about custom storages. @mkabilov had an idea to use odyssey instead of internal proxy in `stolon` so I made that implementation. When client starts working with odyssey it gets connection information (about current master `postgres` instance) from the storage. Then I thought I need some way to reliably cleanup client related data for the storage (each client has some state) and config related data (each storage will have its config too) so I made these memory contexts.

pmwkaa:
In order to integrate this functionality into the project and support your intentions, I propose we need to do following things:

* Instead of having additional library, I suggest those memory pools and context should become part of machinarium
* I would like to avoid touching current memory allocations and model in odyssey for now
* We need to introduce pluggable modules to odyssey where you could implement new functionality and use convenient memory model

This is a big changes and will require some work, but I would love to assist you with this and implement some of the core functionality

g0djan:
Hi, you have a huge list of files with conflicts due to PR with formatting.
Steps for rebasing.
- save files that were changed in your PR
- git checkout master
- git pull
- git checkout <your branch>
- git rebase -i HEAD~0 --onto master(this line delete all your changes, don't forget to save it)
- copy back saved files to the repository
- format files:  find ./ -type f \( -iname \*.c -o -iname \*.h -o -iname \*.cpp -o -iname \*.hpp \) -exec clang-format -style=file -i {} \;
- git commit -am "Name of PR"
- git push -f

If you have a questions or want my help you can ask me here :)

rkhapov:
seems like this PR is abandoned

PR #60, opened 2019-06-25
* Corrected SEGFAULT in case of non-existent user in auth_query

gms80:
+ Added ability for host-based authentification a-la pg_hba:
  e.g.
  auth_query "SELECT usename, passwd FROM pg_shadow,pg_hba_net WHERE usename='%u' AND ((pg_shadow.usename=pg_hba_net.username and network >> '%h') OR (SELECT COUNT(*) FROM pg_hba_net WHERE username='%u'))

### DIFF
--- a/sources/auth.c
+++ b/sources/auth.c
@@ -76,9 +76,11 @@ od_auth_frontend_cleartext(od_client_t *client)
 	kiwi_password_init(&client_password);
 
 	if (client->rule->auth_query) {
+		char peer[128];
+		od_getpeername(client->io.io, peer, sizeof(peer), 1, 0);
 		rc = od_auth_query(client->global,
 		                   client->rule,
-		                   client->io.io,
+		                   peer,
 		                   &client->startup.user,
 		                   &client_password);
 		if (rc == -1) {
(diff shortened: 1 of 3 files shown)


x4m:
@pmwkaa I've fixed formatting, but I think we should reconsider formatting rules: we should have an automatic way to apply formatting.
Btw, maybe refactor that od_getpeername() code in this PR? it is repeated 3 times...

gms80:
I've created another patch, where i pass to the function an IP but not a
"machine_io_t" structure, so od_getpeername is called only once.
Sorry for "space" formatting - thanks to YAML :(



ср, 26 июн. 2019 г. в 08:52, Andrey Borodin <notifications@github.com>:

x4m:
@gms80 what's up with your other patch? should we push this or wait for another PR?

gms80:
Yeah, you can push this.

чт, 11 июл. 2019 г. в 15:09, Andrey Borodin <notifications@github.com>:

gms80:
The second patch commited

gms80:
So, what about to push this PR?

pmwkaa:
Sorry for delay

PR #62, opened 2019-07-23
Make UBSan happy

x4m:


x4m:
According to https://godbolt.org with gcc 5+ generated assembly is identical

PR #66, opened 2019-08-12
WIP: Add pause/resume admin console commands[fork of tomlikestorock's PR]

lowgear:
Original conversation: https://github.com/yandex/odyssey/pull/43

Rebased and fixed build.

### DIFF
--- a/sources/console.c
+++ b/sources/console.c
@@ -27,7 +27,9 @@ enum
 	OD_LCLIENTS,
 	OD_LLISTS,
 	OD_LSET,
-	OD_LPOOLS
+	OD_LPOOLS,
+	OD_PAUSE,
+	OD_RESUME
 };
 
 static od_keyword_t
@@ -41,6 +43,8 @@ od_console_keywords[] =
 	od_keyword("lists",       OD_LLISTS),
 	od_keyword("set",         OD_LSET),
 	od_keyword("pools",       OD_LPOOLS),
+	od_keyword("pause",       OD_PAUSE),
+	od_keyword("resume",      OD_RESUME),
 	{ 0, 0, 0 }
 };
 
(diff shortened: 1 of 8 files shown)


pmwkaa:
Cool, thanks. I'll take a look on the weekend

lowgear:
FYI Work is in progress yet.

sources/console.c
```diff
@@ -793,6 +798,127 @@ od_console_set(od_client_t *client, machine_msg_t *stream)
...
+
+    if (rc == OD_PARSER_KEYWORD)
+    {
+        storage_name = token.value.string.pointer;
+        od_log(&instance->logger, "console", client, NULL,
+                "Making storage %.*s %s...", token.value.string.size, token.value.string.pointer, state_name);
+    }
+    else
+    {
+        od_log(&instance->logger, "console", client, NULL,
```
lowgear: During IRL discussion with @x4m we decided not to implement pause/resume for session pooling. So I am not sure what "for all storages" operation should do. Should it silently apply change only to storages with transaction pooling or report problems with session pooling storages or not work at all if any session pooling storage found?

sources/console.c
```diff
@@ -793,6 +798,127 @@ od_console_set(od_client_t *client, machine_msg_t *stream)
...
+                "Making all storages %s...", state_name);
```
lowgear: During IRL discussion with @x4m we decided not to implement pause/resume for session pooling. So I am not sure what "for all storages" operation should do. Should it silently apply change only to storages with transaction pooling or report problems with session pooling storages or not work at all if any session pooling storage found?

lowgear: After more discussions we decided to implement it. This way there will be no problems in changing pooling mode during puase.

sources/rules.h
```diff
@@ -115,8 +121,31 @@ struct od_rule
...
 	od_list_t               link;
+    /* pause/resume related */
+    pthread_mutex_t lock;
+    od_rule_storage_state_t state;
 };
 
+static inline void
+od_rule_lock(od_rule_t *rule)
+{
+    pthread_mutex_lock(&rule->lock);
```
x4m: We have our internal sleeplock.
It's there for inlining locks.

Actually, I'm not a big fan of bicycles, but, as long as it's already there, let's use it.

pmwkaa:
I am ready to review this, as soon as it is ready, please let me know. I guess it will require to be rebased first.

sources/rules.h
lowgear: I could not find any other lock. And other objects are locked with pthread_mutex_lock too. Could you give a link to certain file/line?

lowgear:
I added reporting of left storages to pause to console during PAUSE command and rebased onto master.

pmwkaa:
@lowgear could you please squash it into one commit?

lowgear:
Squashed my commits

sources/console.c
```diff
@@ -865,54 +882,99 @@ od_console_query_set_storage_state(od_client_t *client, od_parser_t *parser,
...
+    char state_name[] = "PAUSED";
 
     if (rc == OD_PARSER_KEYWORD)
     {
         storage_name = token.value.string.pointer;
         od_log(&instance->logger, "console", client, NULL,
-                "Making storage %s %s...", token.value.string.pointer, state_name);
+                "Making storage %.*s %s...", token.value.string.size, token.value.string.pointer, state_name);
     }
     else
```
pmwkaa: Parsing logic here is a slightly incomplete and have a room for errors. This should expect PARSER_EOF token, otherwise it is possible to pass anything like `PAUSE 1234` and it will treat it as pause all condition.

Also I suggest to follow style convention and start log messages from lowercase letter and remove `...`.

sources/console.c
```diff
@@ -865,54 +882,99 @@ od_console_query_set_storage_state(od_client_t *client, od_parser_t *parser,
...
-        rc = machine_write(client->io, msg);
-        if (rc == -1)
-            return -1;
-        return 0;
+        storage_name = token.value.string.pointer;
+        od_log(&instance->logger, "console", client, NULL,
+               "Making storage %.*s %s...", token.value.string.size, token.value.string.pointer, state_name);
     }
-    if (state == OD_STORAGE_PAUSE)
+    else
```
pmwkaa: Same parsing issues as described above.

lowgear:
I am concerned about integration testing of this feature but I am not familiar with setting up odyssey and posgresql and also setting up testing infrastructure seems like a separate task. If you made such infrastructure I would add testing scenarios for pause/resume.

sources/rules.h
```diff
@@ -122,8 +121,31 @@ struct od_rule
 	int                     client_max;
 	int                     log_debug;
 	od_list_t               link;
+    /* pause/resume related */
+    pthread_mutex_t lock;
```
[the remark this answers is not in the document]
lowgear: I've studied how rules are updated. So I implemented setting pause/resume state the way that circumvent race conditions while config reload is performed.

I don't like the idea of putting state into route because route doesn't represent a DB. Storing state inside a rule isn't good to. The best solution would be to create a new entity that will represent DB's overall state and contain both pause/resume-state and rule(it looks like an atomic part of state). But this is a big change to architecture and should be discussed. Who should I discuss it with? We can discuss it here or I can meet @x4m .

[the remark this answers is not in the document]
pmwkaa: I see no reason for this extension regarding only this change. JFYI statistics is also calculated per-database, you should follow the same logic here I believe.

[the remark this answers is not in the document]
lowgear: It is inconvenient to put pause/resume state into routes. If there is a rule with no routes referencing it pause would make no effect. The best user can get is "no such storage" error but there is such storage. We just can't store pause/resume this way safely.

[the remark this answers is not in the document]
lowgear: @x4m and me discussed that and he approved the idea of adding mutable state object.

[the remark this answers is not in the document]
pmwkaa: I do not see a good reason why you need to implement it, current implementation is optimal for router and other subsystems because you have to support rules where you create automatic routes for `all_database`:`user`, `database`:`all_users`, `all_databases:all_users` as well. It will be harder to maintain it in general and keep up dependencies.

Yet if you brave enough to do it correctly, I would like to see it as a separate PR.

x4m:
Currently, I see two problems:
1. od_strmemcmp is incorrect. null-terminated string can be shorter that other.
2. there is mark field in db state. It is bool, unlike other mark field. Can we make this marks of similar type and some descriptive comments?
3. I think we should wait less that one second to retry routing to paused storage.

x4m:
But, anyway, I propose to merge this as is and fix these problems in separate PRs

PR #68, opened 2019-09-17
Add PAM auth support

reshke:


CMakeLists.txt
```diff
@@ -19,9 +19,9 @@ if ("${CMAKE_BUILD_TYPE}" STREQUAL "")
     set(CMAKE_BUILD_TYPE "Debug")
 endif()
 if ("${CMAKE_BUILD_TYPE}" STREQUAL "Release")
-    set(CMAKE_C_FLAGS "-std=gnu99 -pedantic -Wall -Wextra -Wstrict-aliasing -g -O2")
+    set(CMAKE_C_FLAGS "-std=gnu99 -lpam -pedantic -Wall -Wextra -Wstrict-aliasing -g -O2")
```
pmwkaa: Does including `-lpam` works well with *BSD systems?

odyssey.conf
```diff
@@ -210,7 +210,7 @@ cache_coroutine 0
 #
 # 16KB by default.
 #
-coroutine_stack_size 4
+coroutine_stack_size 8
```
pmwkaa: Is this leftover?

odyssey.conf
```diff
@@ -389,6 +390,10 @@ database default {
 #		auth_query_db ""
 #		auth_query_user ""
 
+#       Authentication PAM.
+#
+#        auth_pam_service "passwd"
```
pmwkaa: Looks like some formatting issues

sources/auth.c
```diff
@@ -75,7 +76,7 @@ od_auth_frontend_cleartext(od_client_t *client)
 	kiwi_password_t client_password;
 	kiwi_password_init(&client_password);
 
-	if (client->rule->auth_query)
+    if (client->rule->auth_query)
```
pmwkaa: Maybe we need some checks in `od_rules_validate` to ensure we do not use auth_query and auth_pam_ together? otherwise it is a bit confusing

odyssey.conf
```diff
@@ -356,6 +356,7 @@ database default {
 #		"clear_text" - PostgreSQL clear text authentication
 #		"md5"        - PostgreSQL MD5 authentication
 #		"cert"       - Compare client certificate Common Name against auth_common_name's
+#		"pam"        - Use PAM service
```
pmwkaa: I do not think this is correct to introduce authentication method here. It should be done the same way as `auth_query` handled and should work both for `clear_text` and `md5` methods, otherwise it gets a bit confusing

sources/auth.c
```diff
@@ -105,6 +106,24 @@ od_auth_frontend_cleartext(od_client_t *client)
 			machine_msg_free(msg);
 			return -1;
 		}
+	} else if(client->rule->auth_pam_service){
```
pmwkaa: This path shuld also be added for `md5` authentication

sources/auth.c
```diff
@@ -344,6 +364,10 @@ int od_auth_frontend(od_client_t *client)
 	case OD_RULE_AUTH_BLOCK:
 		od_auth_frontend_block(client);
 		return -1;
+    case OD_RULE_AUTH_PAM:
```
pmwkaa: This case and `OD_RULE_AUTH_PAM` should be removed as described above to avoid confusion between auth methods methods. PAM should be supported both for clear_text and md5 as well

sources/auth.c
```diff
@@ -13,6 +13,7 @@
 #include <ctype.h>
 #include <inttypes.h>
 #include <assert.h>
+#include <pam.c>
```
pmwkaa: `<pam.c>` ?

sources/pam.c
```diff
@@ -0,0 +1,96 @@
...
+
+    memset(*rspv, 0, msgc * sizeof(struct pam_response));
+    retcode = PAM_SUCCESS;
+
+    for (counter = 0; counter < msgc; counter++) {
+        if (retcode != PAM_SUCCESS)
+            break;
+        switch (msgv[counter]->msg_style) {
+            case PAM_PROMPT_ECHO_OFF:
+                (*rspv)[counter].resp = strdup(request->password);
```
pmwkaa: When this gets freed?

sources/rules.c
```diff
@@ -746,7 +746,9 @@ od_rules_validate(od_rules_t *rules, od_config_t *config, od_logger_t *logger)
 		} else
 		if (strcmp(rule->auth, "cert") == 0) {
 			rule->auth_mode = OD_RULE_AUTH_CERT;
-		} else {
+		} else if (strcmp(rule->auth, "pam") == 0) {
```
pmwkaa: Should be removed, `auth_pam_service` is sufficient

sources/system.c
```diff
@@ -22,6 +22,7 @@
 #include <machinarium.h>
 #include <kiwi.h>
 #include <odyssey.h>
+#include <security/pam_appl.h>
```
pmwkaa: Why do we need it here?

CMakeLists.txt
x4m: AFAIK Machinarium does not work with BSD at all.
PAM seems to be Linux only

pmwkaa: Also, I think this should be removed. Because use `FindPAM` for that.

odyssey.conf
reshke: without this, coroutines do not have enough memory on the stack to call the pam library functions

sources/auth.c
reshke: But the standard protocol postgreSQL only supports cleartext in PAM

sources/pam.c
reshke: This gets freed somewhere deep inside pam library as i understand.

odyssey.conf
```diff
@@ -210,7 +210,7 @@ cache_coroutine 0
 #
 # 16KB by default.
 #
-coroutine_stack_size 4
+coroutine_stack_size 16
```
pmwkaa: I understand the reason, just to be clear that this will increase overall memory consumption per client

sources/pam.c
```diff
@@ -0,0 +1,88 @@
...
+	}
+	
+	memset(*rspv, 0, msgc * sizeof(struct pam_response));
+	retcode = PAM_SUCCESS;
+	
+	for (counter = 0; retcode == PAM_SUCCESS && counter < msgc; counter++) {
+		if (msgv[counter]->msg_style == PAM_PROMPT_ECHO_OFF) {
+			(*rspv)[counter].resp = strdup(request->password);
+			if ((*rspv)[counter].resp == NULL) {
+				retcode = PAM_CONV_ERR;
```
pmwkaa: should we call `break;` here as well?

sources/pam.c
```diff
@@ -0,0 +1,88 @@
...
+	
+	retval = pam_authenticate(pamh, 0);
+	
+	if (retval != PAM_SUCCESS)
+		return retval;
+	
+	retval = pam_acct_mgmt(pamh, PAM_SILENT);
+	
+	if (pam_end(pamh, retval) != PAM_SUCCESS) {     /* close Linux-PAM */
+		pamh = NULL;
```
pmwkaa: why is it here?

sources/pam.c
```diff
@@ -0,0 +1,88 @@
...
+	/* init PAM transaction */
+	retval = pam_start(od_pam_service, user, &conv, &pamh);
+	
+	if (retval != PAM_SUCCESS)
+		return retval;
+	
+	retval = pam_authenticate(pamh, 0);
+	
+	if (retval != PAM_SUCCESS)
+		return retval;
```
pmwkaa: do we need to call `pam_end` here in case of error?

sources/pam.c
```diff
@@ -0,0 +1,88 @@
...
+	
+	if (retcode != PAM_SUCCESS) {
+		for (counter = 0; counter < msgc; counter++)
+			free((*rspv)[counter].resp);
+		free(*rspv);
+	}
+	return retcode;
+}
+
+static int
```
pmwkaa: do we need this function to return PAM statuses, maybe just common `-1, 0` would be more convenient?

sources/pam.c
```diff
@@ -0,0 +1,88 @@
...
+handle_pam_conversation(char *od_pam_service, char * user, char * password){
+	struct auth_request req = {
+		.password=password
+	};
+	struct pam_conv conv = {
+		pam_conversation,
+		.appdata_ptr = &req,
```
pmwkaa: Looks like you do not need this additional `struct auth_request` here and you can pass password pointer as is

sources/pam.c
```diff
@@ -0,0 +1,88 @@
...
+#include <security/pam_appl.h>
+#include <kiwi.h>
+
+struct auth_request {
+    char *password;
+};
+
+
+static int
+pam_conversation(int msgc,
```
pmwkaa: Could you please fix styling in this file until PR with auto-format will be merged

sources/pam.c
reshke: There was sufficient loop condition for that (retcode == PAM_SUCCESS && counter < msgc), but ok, maybe it was bad for readability, so i add if statement to be more explicitly

sources/pam.c
reshke: ok, fixed

sources/pam.c
reshke: fixed

reshke:
Also wrote a little bit in configuration.md 
I hope I provided a satisfactory description

sources/pam.c
reshke: sure, fixed

pmwkaa:
Thanks. Merging this. I've made some refactoring and cleanups worth to mention: do not force dependency on libpam, etc.

PR #70, opened 2019-10-08
Local development inside docker

dimastark:


x4m:
Looks very useful, thanks!

PR #73, opened 2019-11-01
resolve #1: Implements SCRAM Authentication

### LINKED ISSUE #1: Support SCRAM authentication

dimastark:
What needs to be done:
* Channel binding
* Authorization identity
* Mandatory extensions
* Tool for generating md5 and sha256 hashes?
* Remove unnecessary dependency on libpq?


x4m:
Hi! This is a valuable contribution and I'm happy to merge this.
But we need some way to test this and similar functions.

We discussed this possibility with @dimastark in IM and decided that he's going to push some docker tests for this in further PR.

PR #74, opened 2019-11-05
Add client host to default log

reshke:


reshke:
case 'h':
				if (client && client->io.io) {
					od_getpeername(client->io.io, peer, sizeof(peer), 1, 0);
					len = od_snprintf(dst_pos, dst_end - dst_pos, "%s", peer);
					dst_pos += len;
					break;
				}
				len = od_snprintf(dst_pos, dst_end - dst_pos, "none");
				dst_pos += len;
				break;

x4m:
Pushed, thanks!

PR #78, opened 2019-11-22
Add integration tests for SCRAM Authentication

dimastark:


dimastark:
I had to upgrade the ubuntu version in CI to bionic because there is no `postgresql-server-dev-10` package on xenial.

x4m:
Great job, @dimastark ! That's very useful that we can test complex features now.
I think it is better to use the binaries provided by pg_config. I've fixed it temporarily with a flag, but I'll try to do a more portable solution.

x4m:
> real	0m7,644s
This is so fast...
Previous test scripts (in the scripts folder) used to take a lot of time for unclear results.
I'll integrate useful parts of install-check into this infrastructure too.
Thanks!

PR #79, opened 2019-12-01
Optimize clocks

x4m:


third_party/machinarium/sources/clock.c
```diff
@@ -9,14 +9,15 @@
...
 {
-	mm_timer_t *a = *(mm_timer_t**)a_ptr;
-	mm_timer_t *b = *(mm_timer_t**)b_ptr;
 	if (a->timeout == b->timeout) {
+		if (a->seq == b->seq) {
+			assert(a == b);
+			return 0;
+		}
 		return (a->seq > b->seq) ? 1 : -1;
-	} else
```
reshke: As far as i understand we use this in following form in odyssey:)

if {
} else
if

third_party/machinarium/sources/clock.c
```diff
@@ -39,45 +40,81 @@ void mm_clock_free(mm_clock_t *clock)
...
+{
+    int low = 0, high = count;
+    while (low < high)
+    {
+        int med = low + (high - low) / 2;
+        int cmp = mm_clock_cmp(timer, list[med]);
+        if (cmp == 0)
+	        return med;
+        if (cmp > 0) {
+            low = med + 1;
```
mialinx: what about 1 item list with list[med] > timer ? 
med + 1 whould be 1, that's out of list, should be 0 ?

PR #86, opened 2019-12-12
Modifications to build on PG12

erthalion:
Hi,

Thanks for the project! While trying to build it with PostgreSQL 12, I've
stumbled upon couple of issues, namely:

* Incompatibility after [[1]]
* Required presence of libpgport (to expose pg_sprintf for scram implementation)

My first reaction was to introduce the following naive changes (including also
use pg_config to recognise lib/include dir). Without them odyssey, compiled
against v10, was crashing sometimes while connecting to v12. Does something
like that makes sense, or there is a better way to achive compatibility?

[1]: https://git.postgresql.org/gitweb/?p=postgresql.git;a=commit;h=cfc40d384ae51ea2886d599d2008ae57b529e6ea

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,8 +39,18 @@ if (BUILD_DEBIAN)
 endif()
 
 # postgresql
+option(PG_VERSION_NUM "PostgreSQL version" 100000)
 find_package(PostgreSQL REQUIRED)
-set(od_libraries ${od_libraries} ${POSTGRESQL_LIBRARY} ${PQ_LIBRARY})
+execute_process (
+    COMMAND pg_config --libdir
+    OUTPUT_VARIABLE PG_LIBDIR
+)
+execute_process (
+    COMMAND pg_config --includedir-server
+    OUTPUT_VARIABLE PG_INCLUDE_SERVER
+)
+set(POSTGRESQL_INCLUDE_DIR ${PG_INCLUDE_SERVER})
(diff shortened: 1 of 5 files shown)


x4m:
Hi, Dmitry!
This certainly makes sense... something is broking in CI though...
Can you please elaborate a bit on
`Without them odyssey, compiled
against v10, was crashing sometimes while connecting to v12.`
What are conditions to reproduce the crash? Use SCRAM?

erthalion:
> something is broking in CI though...

Yes. With this changes pg_config is expected to be in the path, what is not the case in the tests, but I'll update this. Although locally with pg_config tests were seems to be ok.

> Can you please elaborate a bit on "Without them odyssey, compiled against v10, was crashing sometimes while connecting to v12". What are conditions to reproduce the crash? Use SCRAM?

I haven't investigated that much yet, but I was experimenting with odyssey in a container under alpine. But basically when I configured `storage_user`/`storage_password` I've got segfault. While trying to figure out what happened, I've build odyssey against v12, and this crash disappeared. I can try to get more information on the reason why it had crashed.

erthalion:
> something is broking in CI though...
>
> Yes. With this changes pg_config is expected to be in the path, what is not the case in the tests, but I'll update this. Although locally with pg_config tests were seems to be ok.

I was wrong about this, but eventually the fix was easy.

x4m:
This is a very valuable contribution, thanks!

PR #87, opened 2019-12-15
PAUSE/RESUME fixed

lowgear:


lowgear:
Please, review this :)

sources/config_reader.c
```diff
@@ -1062,6 +1062,15 @@ od_config_reader_import(od_config_t *config, od_rules_t *rules, od_error_t *erro
 		return -1;
 	rc = od_config_reader_parse(&reader);
 	od_config_reader_close(&reader);
+	if (rc == -1)
+		return -1;
+
+	//TODO: We should avoid doing here anything besides importing config.
```
x4m: Please, address this TODO item.
This routine is called on SIGHUP, not only during initialization.

sources/console.c
```diff
@@ -27,7 +27,9 @@ enum
 	OD_LCLIENTS,
 	OD_LLISTS,
 	OD_LSET,
-	OD_LPOOLS
+	OD_LPOOLS,
+	OD_PAUSE,
```
x4m: Please keep prefix ```OD_L``` or discard this L for every other enum member

sources/console.c
```diff
@@ -793,6 +797,228 @@ od_console_set(od_client_t *client, machine_msg_t *stream)
 	return 0;
 }
 
+static inline int
+od_console_route_set_storage_state_cb(od_route_t *route, void **argv)
+{
+	bool pause_all = *(bool*)argv[0];
```
x4m: Can we have struct to stick 5 following params together?

sources/console.c
```diff
@@ -793,6 +797,228 @@ od_console_set(od_client_t *client, machine_msg_t *stream)
...
+	od_router_t *router = client->global->router;
+
+	char *storage_name = NULL;
+	size_t storage_name_size = 0;
+	od_token_t token;
+	int rc = od_parser_next(parser, &token);
+
+	bool all_storages = false;
+
+	switch (rc) {
```
x4m: do we have some logic here that can be reused after pause parsing?

sources/router.c
```diff
@@ -325,6 +328,19 @@ od_router_attach(od_router_t *router, od_config_t *config, od_client_t *client,
 	od_server_t *server;
 	for (;;)
 	{
+		if (!route->rule->db_state->is_active)
```
x4m: Is there any timeout for routing?
Also, we should check wether client disconnected during wait.

lowgear: I will address this a bit later.

lowgear:
I haven't tested the whole thing after issues fixing yet.

sources/router.c
lowgear: I've studied postgess messaging protocol for any timeouts given by the client and haven't found anything. Where should we get routing timeout? Did I miss such a timeout in client's requests or should this timeout by configurable just in the pooler's configs?

lowgear: Talking about waiting I wander if the way I implemented it is correct. At the bottom of the same for cycle some other waiting is implemented by means of od_route_wait. But I'm not sure what and why it actually does.

Also I don't know how to determine if the client is still connected.

lowgear:
Now it's tested that with the given changes pause/resume still works.

g0djan:
Hi, you have a huge list of files with conflicts due to PR with formatting.
Steps for rebasing.

save files that were changed in your PR
git checkout master
git pull
git checkout
git rebase -i HEAD~0 --onto master(this line delete all your changes, don't forget to save it)
copy back saved files to the repository
format files: find ./ -type f ( -iname *.c -o -iname *.h -o -iname *.cpp -o -iname *.hpp ) -exec clang-format -style=file -i {} ;
git commit -am "Name of PR"
git push -f
If you have a questions or want my help you can ask me here :)

rkhapov:
already implemented

PR #89, opened 2019-12-23
Use more hints to find PG libs

x4m:
@erthalion please take a look
Hints from pg_config seem to be not working on Ubuntu installations. Even if variable points to the same path as literal.
I've added literal hint and it helps. But it seems to me we are doing something wrong here.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,17 +39,8 @@ if (BUILD_DEBIAN)
 endif()
 
 # postgresql
-option(PG_VERSION_NUM "PostgreSQL version" 100000)
 find_package(PostgreSQL REQUIRED)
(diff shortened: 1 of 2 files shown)


erthalion:
Oh, interesting. I'll check.

erthalion:
Yep, sorry. I didn't checked it thoroughly enough on Ubuntu (my goal was to check it locally on my Gentoo and on Alpine). You're right, looks like for Ubuntu we also need to check pkglibdir, but also trim results before setting variables. Changes in #90 made it work for me on Ubuntu.

Although it's a bit confusing for me that tests were not complaining for the previous PR. So I want to check it a bit more if everything is in place.

PR #95, opened 2020-01-11
Quantiles for transactions and queries

x4m:
This is a draft for percentile computation.
Usually, the average query time is not very useful, while 99 percentile can give more insights about slow queries.
In this PR I use sampling for query times, this is not a very efficient algorithm like T-digest.
I think in future we will implement something more precise.

### DIFF
--- a/odyssey.conf
+++ b/odyssey.conf
@@ -500,7 +500,7 @@ database default {
 #
 		log_debug no
 
-#       Compute quantiles of query and transaction times
+#		Compute quantiles of query and transaction times
 		quantiles "0.99,0.95,0.5"
 	}
 }
--- a/odyssey.conf
+++ b/odyssey.conf
@@ -499,6 +499,9 @@ database default {
 #		Enable verbose mode for a specific route only.
 #
 		log_debug no
+
+#       Compute percentiles of query and transaction times
+		percentiles "0.99,0.95,0.5"
 	}
 }
 
--- a/odyssey.conf
+++ b/odyssey.conf
@@ -500,8 +500,8 @@ database default {
 #
 		log_debug no
 
-#       Compute percentiles of query and transaction times
-		percentiles "0.99,0.95,0.5"
+#       Compute quantiles of query and transaction times
+		quantiles "0.99,0.95,0.5"
 	}
 }
 
--- a/sources/config_reader.c
+++ b/sources/config_reader.c
@@ -93,7 +93,8 @@ enum
 	OD_LAUTH_PAM_SERVICE,
 	OD_LAUTH_QUERY,
 	OD_LAUTH_QUERY_DB,
-	OD_LAUTH_QUERY_USER
+	OD_LAUTH_QUERY_USER,
+    OD_LPERCENTILES,
 };
 
 typedef struct
@@ -182,6 +183,7 @@ od_config_keywords[] =
 	od_keyword("auth_query_db",        OD_LAUTH_QUERY_DB),
 	od_keyword("auth_query_user",      OD_LAUTH_QUERY_USER),
 	od_keyword("auth_pam_service",     OD_LAUTH_PAM_SERVICE),
+    od_keyword("percentiles",          OD_LPERCENTILES),
 	{ 0, 0, 0 }
 };
 
@@ -292,6 +294,36 @@ error:
 	od_parser_push(&reader->parser, &token);
 	od_config_reader_error(reader, &token, "expected '%c'", symbol);
 	return false;
+}
+
+static bool
+od_config_reader_percentiles(od_config_reader_t *reader, char *value, double **percentiles, int *count)
+{
+    int comma_cnt = 1;
+    char *c = value;
+    while (*c){
+        if (*c == ',')
+            comma_cnt++;
+        c++;
+    }
+    *percentiles = malloc(sizeof(double) * comma_cnt);
+    double *array = *percentiles;
+    *count = 0;
+    c = value;
+    while (*c) {
+        int length = sscanf(c, "%lf", array + *count);
+        if (length !=1 || array[*count] > 1 ||array[*count] < 0) {
+            od_config_reader_error(reader, NULL, "incorrect percentile value");
+            free(*percentiles);
+            return false;
+        }
+        *count += 1;
+        while (*c != ',') {
+            c++;
+        }
+        c++;
+    }
+    return true;
 }
 
 static bool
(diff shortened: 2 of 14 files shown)


sources/config_reader.c
```diff
@@ -294,6 +296,36 @@ od_config_reader_symbol(od_config_reader_t *reader, char symbol)
 	return false;
 }
 
+static bool
+od_config_reader_percentiles(od_config_reader_t *reader, char *value, double **percentiles, int *count)
```
g0djan: And all this code just for splitting a string  to a double array.

g0djan: Also, it's possible to use strsep http://man7.org/linux/man-pages/man3/strsep.3.html
But I guess it will not really simplify this function

sources/config_reader.c
```diff
@@ -693,6 +725,17 @@ od_config_reader_route(od_config_reader_t *reader, char *db_name, int db_name_le
 			if (! od_config_reader_yes_no(reader, &route->client_fwd_error))
 				return -1;
 			continue;
+		/* percentiles */
+		case OD_LPERCENTILES:
+        {
```
g0djan: Indentation

sources/config_reader.c
g0djan: It will be better to return double* array(but as I see it's a common pattern here) 
or at least pass it as double* instead of double**

sources/cron.c
```diff
@@ -92,6 +92,21 @@ od_cron_stat_cb(od_route_t *route, od_stat_t *current, od_stat_t *avg,
 	       info.avg_recv_client,
 	       info.avg_recv_server);
 
+	for (int i = 0; i < route->rule->percentiles_count; i++) {
+        double perc = route->rule->percentiles[i];
```
g0djan: Is it https://www.urbandictionary.com/define.php?term=percs ?
Use full word

sources/cron.c
```diff
@@ -92,6 +92,21 @@ od_cron_stat_cb(od_route_t *route, od_stat_t *current, od_stat_t *avg,
...
+        uint64_t qp = 0;
```
g0djan: query_percentile instead of qp?

sources/cron.c
```diff
@@ -92,6 +92,21 @@ od_cron_stat_cb(od_route_t *route, od_stat_t *current, od_stat_t *avg,
...
+        uint64_t tp = 0;
```
g0djan: transaction_percentile instead of tp?

sources/hgram.h
```diff
@@ -0,0 +1,28 @@
...
+#define OD_HGRAM_DATA_POINTS 256
+
+struct od_hgram {
+	uint32_t data[OD_HGRAM_DATA_POINTS];
+	uint64_t estimated_size;
+};
+
+void od_hgram_init(od_hgram_t *);
+
+void od_hgram_add_data_point(od_hgram_t *, uint64_t);
```
g0djan: Would it be useful to return a bool and know was it successful adding or not?

sources/rules.h
```diff
@@ -116,6 +116,8 @@ struct od_rule
 	int                     client_max_set;
 	int                     client_max;
 	int                     log_debug;
+	double 				   *percentiles;
+	int 				    percentiles_count;
```
g0djan: Indentation

sources/stat.h
```diff
@@ -72,6 +78,8 @@ od_stat_query_end(od_stat_t *stat, od_stat_state_t *state,
 		if (diff > 0) {
 			od_atomic_u64_add(&stat->tx_time, diff);
 			od_atomic_u64_inc(&stat->count_tx);
+            if (stat->transaction_hgram)
+                od_hgram_add_data_point(stat->transaction_hgram, diff);
```
g0djan: Indentation

sources/stat.h
```diff
@@ -24,6 +26,8 @@ struct od_stat
 	od_atomic_u64_t tx_time;
 	od_atomic_u64_t recv_server;
 	od_atomic_u64_t recv_client;
+	od_hgram_t*      transaction_hgram;
+	od_hgram_t*      query_hgram;
```
g0djan: Indentation

test/odyssey/test_hgram.c
```diff
@@ -0,0 +1,91 @@
...
+
+	for (int i = 0; i < 100; i++) {
+		od_hgram_add_data_point(&hgram, 100 - i);
+	}
+
+	od_hgram_freeze(&hgram, &f);
+
+	assert(od_hgram_percentile(&f, 0.7) == 70);
+	assert(od_hgram_percentile(&f, 0.5) == 50);
+	assert(od_hgram_percentile(&f, 0.3) == 30);
```
g0djan: 70'th percentile should be 30, 30'th percentile should be 70 for a definition from https://en.wikipedia.org/wiki/Percentile

test/odyssey/test_hgram.c
```diff
@@ -0,0 +1,91 @@
...
+#include <lrand48.h>
+#include "sources/hgram.h"
+
+void hgram_forward_test();
+
+void hgram_backward_test();
+
+int hgram_random_test();
+
+static int myround(double x)
```
g0djan: http://www.cplusplus.com/reference/cmath/round/ round have the same meaning

odyssey.conf
```diff
@@ -499,6 +499,9 @@ database default {
 #		Enable verbose mode for a specific route only.
 #
 		log_debug no
+
+#       Compute quantiles of query and transaction times
+		quantiles "0.99,0.95,0.5"
```
g0djan: Indentation

sources/config_reader.c
g0djan: And nothing happened? Fix it, pls

sources/stat.h
g0djan: Fix it

g0djan:
And it will be better to mark resolved issues by resolve button

sources/config_reader.c
```diff
@@ -294,6 +296,36 @@ od_config_reader_symbol(od_config_reader_t *reader, char symbol)
 	return false;
 }
 
+static bool
+od_config_reader_percentiles(od_config_reader_t *reader, char *value, double **percentiles, int *count)
+{
+    int comma_cnt = 1;
+    char *c = value;
```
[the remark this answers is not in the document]
x4m: In code style adopted in Odyssey, we used to right pointer alignment. BTW it's quite common pattern in C projects, but PG use left alignment. Probably, one day we will use it too...

sources/config_reader.c
x4m: It's common pattern in config parsing, I don't think to rewrite it fits into this PR.

sources/config_reader.c
x4m: Oh, seems like I missed it, sorry

sources/stat.h
x4m: hmmm, i used wrong pointer allignment here...

x4m:
@g0djan thanks for reviewing this.
BTW I think we should just move to automatic code formatting, so tedious indentation reviewing is not strictly necessary.

g0djan:
Okay, but I heard about it several times and it still has no formatter or linter.

x4m:
We are going to merge #69 right after #87 : I do not want reformatting to be on a way of PAUSE.

test/odyssey/test_hgram.c
x4m: Currently, we do not link with math library. It was common policy for a while. Eventually, we will change that, but not in this PR.

g0djan:
Approved

PR #98, opened 2020-01-17
Remove useless logging of startup errors

x4m:
Currently, logs are flooded with messages like ```msg=startup packet read error``` from pings of monitoring and HA tools.
This commit removes those lines.

### DIFF
--- a/sources/cron.h
+++ b/sources/cron.h
@@ -13,6 +13,7 @@ struct od_cron
 {
 	uint64_t     stat_time_us;
 	od_global_t *global;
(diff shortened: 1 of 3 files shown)


sources/cron.h
```diff
@@ -13,6 +13,7 @@ struct od_cron
 {
 	uint64_t     stat_time_us;
 	od_global_t *global;
+	od_atomic_u64_t startup_errors;
```
reshke: we can make this operation non-atomic since the exact number of errors of this kind is not very important

PR #107, opened 2020-01-27
better replication support

efimkin:


x4m:
This seems to break existing storage setting, but also make them much more straightforward.
LGTM. Thanks!

PR #122, opened 2020-02-19
Fix percentile read

x4m:


sources/config_reader.c
```diff
@@ -327,9 +327,11 @@ od_config_reader_quantiles(od_config_reader_t *reader, char *value, double **qua
             return false;
         }
         *count += 1;
-        while (*c != ',') {
+        while (*c != 0 && *c != ',') {
```
g0djan: Hmm, it looks like you tried to check that "c" is nullptr but compare value "*c" instead of pointer "c".

x4m: I've fixed that

PR #126, opened 2020-02-24
Fix pthread

x4m:


x4m:
I think that appropriate fix was already merged as part of #128

PR #131, opened 2020-03-05
Do not use abort() in frontend code

x4m:
While this is certainly "can't happen" error, it is not unrecoverable yet.

### DIFF
--- a/sources/frontend.c
+++ b/sources/frontend.c
@@ -881,7 +881,9 @@ od_frontend_cleanup(od_client_t *client, char *context,
 		break;
 
 	default:
-		abort();
+        od_error(&instance->logger, context, client, server,
+                 "%s", "unexpected error status %s (%d)", od_status_to_str(status), (uint32)status);
+        od_router_close(router, client);
 		break;
 	}
 }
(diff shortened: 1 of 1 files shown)


sources/frontend.c
```diff
@@ -881,7 +881,9 @@ od_frontend_cleanup(od_client_t *client, char *context,
 		break;
 
 	default:
```
pmwkaa: I think that using abort is more appropriate here, because this condition could only happen in case of programming error and not related to the protocol.

x4m: Yup, I have 6 core dumps.
But they are useless: there is not enough information.
Appropriate logging would be much better.

x4m: I've removed handling of corrupted statuses with attempt of logging. @pmwkaa can you plz look into this again?

g0djan: I agree that it would be better to fix the root of problem, but it will be easier to find roots of problems with logging. So LGTM

x4m: Finally, that was "unexpected error status OD_DETACH (4)". But how could we find this out without logs?

PR #139, opened 2020-04-01
log_client_peer_name_on_error

reshke:


x4m:
I think that in case of error we should always provide all the relevan information unless it's excesive. So, I think that proposed change is fine and should be always on.

PR #141, opened 2020-04-02
add peername to erros log

reshke:


sources/frontend.c
```diff
@@ -1007,44 +1007,67 @@ od_frontend(void *arg)
...
-		         client->startup.database.value,
-		         client->startup.user.value);
-		od_frontend_error(client, KIWI_UNDEFINED_DATABASE,
-		                  "route for '%s.%s' is not found",
-		                  client->startup.database.value,
-		                  client->startup.user.value);
-		od_frontend_close(client);
-		return;
+	{
+        char peer[128];
```
x4m, line 1011: Can we somehow avoid duplicating this code so many times?

PR #147, opened 2020-05-13
t-digest for latencies

g0djan:
This algorithm has an experiment proves high efficiency. We can support it to have the ability to calculate quantiles even for high-rate "SELECT 1" queries.
I found only one serious trouble for using it - it doesn't support the calculation of quantiles in the rolling window by default.

But it has a cool property - it's possible to merge t-digests.
So there is a solution:
- store several t-digests
- to get a quantile - merge several t-digests into one, and ask quantile
- to drop old values - drop the oldest t-digest and start a new one


g0djan:
@x4m if you have ideas or requests for this feature you can put it here.

sources/stat.h
```diff
@@ -20,14 +21,16 @@ struct od_stat_state
...
 	od_atomic_u64_t count_query;
 	od_atomic_u64_t count_tx;
 	od_atomic_u64_t query_time;
 	od_atomic_u64_t tx_time;
 	od_atomic_u64_t recv_server;
 	od_atomic_u64_t recv_client;
-	od_hgram_t *transaction_hgram;
-	od_hgram_t *query_hgram;
+	td_histogram_t* transaction_hgram[5];
+    td_histogram_t* query_hgram[5];
```
x4m, line 33: Why 5?

sources/stat.h
```diff
@@ -65,8 +68,10 @@ od_stat_query_end(od_stat_t *stat,
 			*query_time = diff;
 			od_atomic_u64_add(&stat->query_time, diff);
 			od_atomic_u64_inc(&stat->count_query);
-			if (stat->query_hgram)
-				od_hgram_add_data_point(stat->query_hgram, diff);
+            if (stat->enable_quantiles) {
+                td_add(stat->query_hgram[stat->current_tdigest], diff, 1);
+            }
+//				od_hgram_add_data_point(stat->query_hgram, diff);
```
x4m, line 74: extraneous comment

sources/router.c
```diff
@@ -268,10 +269,10 @@ od_router_route(od_router_t *router, od_config_t *config, od_client_t *client)
 		is_shared = od_config_is_multi_workers(config);
 		route = od_route_pool_new(&router->route_pool, is_shared, &id, rule);
 		if (route == NULL) {
-			od_router_unlock(router);
-			return OD_ROUTER_ERROR;
-		}
-	}
+            od_router_unlock(router);
```
x4m, line 272: I though we have automatic formatting...

sources/router.h
```diff
@@ -51,14 +51,20 @@ od_router_reconfigure(od_router_t *, od_rules_t *);
 int
 od_router_expire(od_router_t *, od_list_t *);
 void
-od_router_gc(od_router_t *);
+od_router_gc(od_router_t *router);
```
x4m, line 54: Well, we did not have names in prototypes. I don't think it is right. But let's keep things alligned - either all with names, or all without.

sources/tdigest.c
```diff
@@ -0,0 +1,322 @@
...
+
+    node_t nodes[0];
+};
+
+static bool is_very_small(double val) {
+    return !(val > .000000001 || val < -.000000001);
+}
+
+static int cap_from_compression(double compression) {
+    return (6 * (int)(compression)) + 10;
```
x4m, line 39: a comment what is going on here would be helpful...

sources/tdigest.c
```diff
@@ -0,0 +1,322 @@
+#include <stdlib.h>
+#include <stdbool.h>
+#include <string.h>
+#include <math.h>
+
+#include "tdigest.h"
+
+#define M_PI 3.14159265358979323846
```
x4m, line 8: what's wrong with math.h's Pi? Not enough digits? Don't we need double suffix here?

test/odyssey/test_tdigest.c
```diff
@@ -0,0 +1,115 @@
...
+
+    double quantiles[3] = {0.5, 0.9, 0.99};
+    for (size_t i = 0; i < 3; ++i) {
+        assert(td_value_at(common_hist, quantiles[i]) != td_value_at(hists[0], quantiles[i]));
+    }
+}
+
+void tdigest_test(void) {
+    simple_test();
+    monotonicity_test();
```
x4m, line 111: plz apply formatting, you merged it for goo reason :)

third_party/machinarium/sources/write.c
```diff
@@ -8,6 +8,8 @@
 #include <machinarium.h>
 #include <machinarium_private.h>
 
+#define IOV_MAX __IOV_MAX
```
x4m, line 11: How is it related?

x4m:
Thanks for working on this! It looks like promising technique, better than Vitter's reservoir.

sources/stat.h
g0djan: That was in start version. I thought that 5 minutes is pretty ok. I can replace it with setting from config.

sources/router.c
g0djan: I forgot about it)

sources/tdigest.c
g0djan: It will keep intrigued until the next PR)

third_party/machinarium/sources/write.c
g0djan: Spare

deploy
```diff
@@ -0,0 +1,3 @@
+rm /usr/bin/odyssey
```
x4m, line 1: This script seems to lack extension and portability... do we really want it in main repo?

sources/CMakeLists.txt
```diff
@@ -26,6 +26,7 @@ set(od_src
     hgram.c
```
x4m, line 26: Don't we want to remove hgram code? It's obsolete now?

sources/stat.h
```diff
@@ -65,8 +70,9 @@ od_stat_query_end(od_stat_t *stat,
 			*query_time = diff;
 			od_atomic_u64_add(&stat->query_time, diff);
 			od_atomic_u64_inc(&stat->count_query);
-			if (stat->query_hgram)
-				od_hgram_add_data_point(stat->query_hgram, diff);
+			if (stat->enable_quantiles) {
+				td_add(stat->query_hgram[stat->current_tdigest], diff, 1);
```
x4m, line 74: Is it thread safe? Will it work with concurrent merges?

test/odyssey/test_tdigest.c
```diff
@@ -0,0 +1,128 @@
...
+	for (size_t i = 0; i < 3; ++i) {
+		assert(td_value_at(common_hist, quantiles[i]) !=
+		       td_value_at(hists[0], quantiles[i]));
+	}
+}
+
+void
+tdigest_test(void)
+{
+	simple_test();
```
x4m, line 123: Can we also add concurrency test?

sources/stat.h
g0djan: It isn't thread-safe. I fix it with the same trick as you - using memcpy.

sources/CMakeLists.txt
g0djan: Deleted

deploy
g0djan: Deleted

test/odyssey/test_tdigest.c
g0djan: It isn't thread-safe. So there is no necessary to concurrency test.

test/odyssey/test_tdigest.c
```diff
@@ -112,11 +114,97 @@ merge_several_digests_test()
...
+	td_histogram_t *freeze    = td_new(100);
+
+	for (int i = 0; i < 100000; i++) {
+		td_add(histogram, machine_lrand48() % 10000, 1);
+	}
+
+	td_copy(freeze, histogram);
+
+	int result = 0;
+	if (myround(td_value_at(freeze, 0.8) / 2000.0) != 4)
```
x4m, line 163: when you linked with m there is no more point in myround...

PR #151, opened 2020-06-01
Fix memory leak at auth.c

lowgear:
free pointer after use

The issue was detected by covertry:
https://scan8.coverity.com/reports.htm#v40721/p12618/fileInstanceId=32935024&defectInstanceId=10694028&mergedDefectId=286530&eventId=10694028-68


lowgear:
I ran tests locally and found that both in master and PR-branch tests pass in release build and fail in debug.

lowgear:
But turned out there is change in release-clang configuration. master passes and PR-branch fails.

Issue #138, opened 2020-04-01
Show user in connection error message


georgthegreat:
If wrong username was passed in connection string, Odyssey will respond with the following error message:
```
ERROR:  odyssey: cd0cc8a7b6adf: user blocked
```

It would be nice to have username given in this message. This could simplify debugging in some cases.

solozobov:
continuing idea of @georgthegreat
Also I think it would be better if we distinguish error messages of three kinds:
1. "unknown user '%username%'", when wrong username was passed in connection string
2. "user '%username%' is not authorized to connect to database '%database%'" when user don't have permission to connect to specified database

x4m:
Also, in the same task we need to improve messages about connection limits:
1. When overall connection limit reached
2. When pool connection limit reached
3. When active connections limit reached

Currently, errors vary in politeness and punctuation. We need more descriptive and clear messages here.

nickitat:
Kind of what @solozobov mentioned in the second point. If you try to connect to non-existing database you will get the same "user blocked" error message.

x4m:
@nickitat @georgthegreat @solozobov Can I ask you to review error messages in linked pull request? Just If you have some spare time..

PR #153, opened 2020-06-02
Improve frequent user error messages

x4m:
Address #138


sources/frontend.c
```diff
@@ -881,7 +881,7 @@ od_frontend_cleanup(od_client_t *client, char *context, od_status_t status)
 			assert(client->route != NULL);
 			od_frontend_error(client,
 			                  KIWI_TOO_MANY_CONNECTIONS,
-			                  "sorry, too many clients for pool");
+			                  "too many active clients for user (pool_size for user reached)");
```
georgthegreat, line 884: could we provide user_name and pool_size here?

x4m: Thanks, I've addressed this issue

georgthegreat: LGTM

PR #157, opened 2020-06-03
do teardown if setup or any test failed

lowgear:
If any test failed abort would be called so teardown is skipped and postgresql remains up.

Here we do teardown if setup or a test failed.


test/odyssey_test.h
```diff
@@ -26,17 +26,17 @@
 		          (tv_stop.tv_usec - tv_start.tv_usec) / 1000);                \
 	} while (0);
 
-#define odyssey_shell_test(file)                                               \
+#define odyssey_shell_test(file, rc, on_fail)                                  \
```
g0djan, line 29: You passed rc by value.

lowgear: that is a macro

PR #171, opened 2020-06-14
check style only in specific folders

lowgear:


g0djan:
I hope it will fix master

PR #177, opened 2020-06-15
Update ubuntu image version

heckad:
The old ubuntu version uses an old cmake which didn't support `add_compile_definitions`


x4m:
@lowgear is this ```add_compile_definitions``` really widely used functionality?
I had to update my cmake to kitware's too.

x4m:
Well, updating docker file won't hurt anyway.
Thanks, @heckad !

PR #183, opened 2020-06-25
Add parse pgoptions and extract search_path

heckad:
Possible implementation parser of pgoptions vars.

Need to do:
fix update argument if they have not changed
add tests
refactor

### DIFF
--- a/third_party/kiwi/kiwi/be_read.h
+++ b/third_party/kiwi/kiwi/be_read.h
@@ -64,6 +64,8 @@ kiwi_be_read_options(kiwi_be_startup_t *su,
 			kiwi_var_set(&su->database, KIWI_VAR_UNDEF, value, value_size);
 		else if (name_size == 12 && !memcmp(name, "replication", 12))
 			kiwi_var_set(&su->replication, KIWI_VAR_UNDEF, value, value_size);
+		else if (name_size == 8 && !memcmp(name, "options", 8))
+			kiwi_parse_pgoptions_and_update_vars(vars, value);
 		else
 			kiwi_vars_update(vars, name, name_size, value, value_size);
 	}
--- a/third_party/kiwi/kiwi/be_read.h
+++ b/third_party/kiwi/kiwi/be_read.h
@@ -7,6 +7,8 @@
  * postgreSQL protocol interaction library.
  */
 
+#include "parser_pgoptions.h"
+
 typedef struct kiwi_be_startup kiwi_be_startup_t;
 
 struct kiwi_be_startup
@@ -65,7 +67,7 @@ kiwi_be_read_options(kiwi_be_startup_t *su,
 		else if (name_size == 12 && !memcmp(name, "replication", 12))
 			kiwi_var_set(&su->replication, KIWI_VAR_UNDEF, value, value_size);
 		else if (name_size == 8 && !memcmp(name, "options", 8))
-			kiwi_parse_pgoptions_and_update_vars(vars, value);
+			kiwi_parse_pgoptions_and_update_vars(vars, value, value_size);
 		else
 			kiwi_vars_update(vars, name, name_size, value, value_size);
 	}
--- a/third_party/kiwi/kiwi/parser_pgoptions.c
+++ b/third_party/kiwi/kiwi/parser_pgoptions.c
@@ -40,13 +40,23 @@ ParseLongOption(const char *string, char **name, char **value)
 }
 
 int
-kiwi_parse_pgoptions_and_update_vars(kiwi_vars_t *vars, char *pgoptions)
+kiwi_parse_pgoptions_and_update_vars(kiwi_vars_t *vars,
+                                     char *pgoptions,
+                                     int pgoptions_len)
 {
 	int errs = 0;
 	int flag;
 
 	int argc;
 	char **argv;
+	char *old_pgoptions = NULL;
+
+	if (pgoptions[pgoptions_len - 1] != '\0') {
+		char *old_pgoptions = pgoptions;
+		pgoptions           = malloc(pgoptions_len + 1);
+		memcpy(pgoptions, old_pgoptions, pgoptions_len);
+		pgoptions[pgoptions_len] = '\0';
+	}
 
 	wordexp_t res;
 	wordexp(pgoptions, &res, 0);
@@ -81,6 +91,10 @@ kiwi_parse_pgoptions_and_update_vars(kiwi_vars_t *vars, char *pgoptions)
 	}
 
 	optind = 1;
+
+	if (old_pgoptions) {
+		free(pgoptions);
+	}
 
 	return errs;
 }
--- a/third_party/kiwi/kiwi/parser_pgoptions.c
+++ b/third_party/kiwi/kiwi/parser_pgoptions.c
@@ -1,0 +1,86 @@
+
+/*
+ * kiwi.
+ *
+ * postgreSQL protocol interaction library.
+ */
+
+#include <getopt.h>
+#include <wordexp.h>
+#include <assert.h>
+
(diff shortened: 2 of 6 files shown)


third_party/kiwi/kiwi/be_read.h
```diff
@@ -64,6 +64,8 @@ kiwi_be_read_options(kiwi_be_startup_t *su,
 			kiwi_var_set(&su->database, KIWI_VAR_UNDEF, value, value_size);
 		else if (name_size == 12 && !memcmp(name, "replication", 12))
 			kiwi_var_set(&su->replication, KIWI_VAR_UNDEF, value, value_size);
+		else if (name_size == 8 && !memcmp(name, "options", 8))
+			kiwi_parse_pgoptions_and_update_vars(vars, value);
```
x4m, line 68: We do not pass value_size here, but value have a semantics of buffer, not a string. Coincidentally there may be \0 byte. But it's not necessary to be this way.

third_party/kiwi/kiwi/var.h
```diff
@@ -99,10 +125,35 @@ kiwi_vars_init(kiwi_vars_t *vars)
...
+	                        "application_name",
+	                        17,
+	                        KIWI_VAR_APPLICATION_NAME);
+	kiwi_var_init_with_type(&vars->vars[KIWI_VAR_SEARCH_PATH],
+	                        "search_path",
+	                        12,
+	                        KIWI_VAR_SEARCH_PATH);
 }
 
+// static inline void
```
x4m, line 138: plz remove commented code

third_party/kiwi/kiwi/var.h
```diff
@@ -99,10 +125,35 @@ kiwi_vars_init(kiwi_vars_t *vars)
 	kiwi_var_init(&vars->vars[KIWI_VAR_STANDARD_CONFORMING_STRINGS],
 	              "standard_conforming_strings",
 	              28);
-	kiwi_var_init(
-	  &vars->vars[KIWI_VAR_APPLICATION_NAME], "application_name", 17);
+	kiwi_var_init_with_type(&vars->vars[KIWI_VAR_APPLICATION_NAME],
+	                        "application_name",
+	                        17,
+	                        KIWI_VAR_APPLICATION_NAME);
+	kiwi_var_init_with_type(&vars->vars[KIWI_VAR_SEARCH_PATH],
```
x4m, line 132: do you actually need it with type? I think at this point client has no preferences about search path and app name..

heckad: As I understand `od_deploy` is the main synchronization method between client configuration and current server connection. If we disconnect from the odyssey and try to connect with a new app name then in this method all values (from client and server) will be compared and if will find different than will build a query for change parameters on the server. Iteration goes through `kiwi_var_type_t` enum because need to add new var to this enum.

third_party/kiwi/kiwi/be_read.h
heckad: Is it possible to have a string without `'\0'` at the end? I use `wordexp` function wich needed for `'\0'`.

x4m: You can copy memory to palloc-ed area with +1 extra byte.

heckad: What is the best way to use `palloc` with kiwi? If I add `#include "postgres.h"` I got an error: `fatal error: postgres.h: No such file or directory`

x4m: Sorry, I'm a bit biased toward pg. Just use malloc().

heckad: Done

third_party/kiwi/kiwi/var.h
x4m: Yes, but why not use kiwi_var_init()? As far as I can see you change application_name behavior and I want to understand if something is wrong currently.

third_party/kiwi/kiwi/parser_pgoptions.c
```diff
@@ -0,0 +1,100 @@
...
+		memcpy(pgoptions, old_pgoptions, pgoptions_len);
+		pgoptions[pgoptions_len] = '\0';
+	}
+
+	wordexp_t res;
+	wordexp(pgoptions, &res, 0);
+
+	argc = res.we_wordc + 1;
+	argv = malloc(sizeof(char *) * (argc + 1));
+	if (argv == 0) {
```
x4m, line 66: It's more ideomatic to compare ptr with NULL

third_party/kiwi/kiwi/parser_pgoptions.c
```diff
@@ -0,0 +1,100 @@
...
+		return -1;
+	}
+
+	argv[0] = "fake_word"; // needs for getopt function
+
+	memcpy(&argv[1], res.we_wordv, sizeof(char *) * argc);
```
x4m, line 72: Do not forget to place that 0 byte at the end.

third_party/kiwi/kiwi/var.h
```diff
@@ -46,6 +47,18 @@ kiwi_var_init(kiwi_var_t *var, char *name, int name_len)
 	var->value_len = 0;
 }
 
+static inline void
+kiwi_var_init_with_type(kiwi_var_t *var,
+                        char *name,
+                        int name_len,
+                        kiwi_var_type_t type)
+{
+	var->type      = type;
```
x4m, line 56: Is this type necessary?

third_party/kiwi/kiwi/parser_pgoptions.c
heckad: `res.we_wordv` yet contains 0 at the end.

third_party/kiwi/kiwi/var.h
heckad: Yes, without the correct type `kiwi_vars_cas` don't work correctly. When `kiwi_vars_cas` called `kiwi_var_compare` type mismatch happens.

third_party/kiwi/kiwi/parser_pgoptions.c
```diff
@@ -0,0 +1,100 @@
...
+	int errs = 0;
+	int flag;
+
+	int argc;
+	char **argv;
+	char *old_pgoptions = NULL;
+
+	if (pgoptions[pgoptions_len - 1] != '\0') {
+		char *old_pgoptions = pgoptions;
+		pgoptions           = malloc(pgoptions_len + 1);
```
x4m, line 56: here malloc can be unsuccessful, i.e. return NULL

third_party/kiwi/kiwi/parser_pgoptions.c
```diff
@@ -0,0 +1,100 @@
...
+		pgoptions[pgoptions_len] = '\0';
+	}
+
+	wordexp_t res;
+	wordexp(pgoptions, &res, 0);
+
+	argc = res.we_wordc + 1;
+	argv = malloc(sizeof(char *) * (argc + 1));
+	if (argv == NULL) {
+		return -1;
```
x4m, line 67: pgoption can leak here

x4m:
Sorry, I've completely forgot about this PR. I think we should get it merged eventually. This PR is marked as WiP, but as far as I remember we decided that it's viable to commit it as is, just fix linting. Am I right?

pvlltvk:
This would be a very useful feature for us because `psycopg2` uses `options` to set  `search_path` for a session.

x4m:
@heckad plz rebase the patch and let's merge this. There are so many requests for the feature already...

heckad:
https://github.com/yandex/odyssey/pull/340

PR #193, opened 2020-07-09
Protection from broken sync during packet pipelining

x4m:
Segfault can be reproduced by this go code

```
package main

import (
        "fmt"
        "github.com/jackc/pgproto3"
)
import "os"
import "context"
import pgx "github.com/jackc/pgx"

func main() {
        ctx := context.Background()
        conn, err := pgx.Connect(ctx, "host=::1 port=6432 user=x4mmm database=postgres")
        if err != nil {
                fmt.Fprintf(os.Stderr, "Unable to connection to database: %v\n", err)
                os.Exit(1)
        }
        pgConn := conn.PgConn().Conn()
        buf := make([]byte,8192)
        buf = (&pgproto3.Query{String: "select 1"}).Encode(buf)
        buf[0] = 0x80
        buf[1] = 0x80
        buf[2] = 0x80
        buf[3] = 0x80
        pgConn.Write(buf)
}

```

### DIFF
--- a/sources/frontend.c
+++ b/sources/frontend.c
@@ -971,6 +971,7 @@ od_frontend_cleanup(od_client_t *client,
 		case OD_SKIP:
 		case OD_ATTACH:
 		case OD_DETACH:
+		case OD_ESYNC_BROKEN:
 			od_error(&instance->logger,
 			         context,
 			         client,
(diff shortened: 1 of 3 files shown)


reshke:
very cool

PR #195, opened 2020-07-15
Fix incomplete odyssey err logging

reshke:


x4m:
Let's merge this for now.
But. Our logging system is a mess. Let's avoid passing loggers here and there in future. Just make a logger part of a global state.

PR #197, opened 2020-07-21
tcp keepalive settings

reshke:


x4m:
Cool! Maybe also add ```tcp_user_timeout``` here too?

odyssey.conf
```diff
@@ -226,6 +226,18 @@ nodelay yes
 #
 keepalive 7200
```
x4m, line 227: Let's make this number as deafult 15 seconds?

PR #208, opened 2020-09-02
show errors per route

reshke:


sources/console.c
```diff
@@ -341,6 +333,151 @@ od_console_show_errors(od_client_t *client, machine_msg_t *stream)
...
+
+			char data[64];
+			int data_len;
+			data_len = od_snprintf(data, sizeof(data), "%" PRIu64, total_count);
+
+			rc = kiwi_be_write_data_row_add(stream, offset, data, data_len);
+			if (rc != OK_RESPONSE) {
+				return rc;
+			}
+			/*            */
```
x4m, line 396: Zen comment

odyssey-dev.conf
```diff
@@ -75,9 +75,39 @@ database default {
...
+
+		client_fwd_error yes
+
+		application_name_add_host yes
+
+		server_lifetime 3600
+		log_debug no
+
+		quantiles "0.99,0.95,0.5"
+        client_max 107
```
x4m, line 108: ill-formatted dev config

sources/console.c
```diff
@@ -341,6 +333,151 @@ od_console_show_errors(od_client_t *client, machine_msg_t *stream)
 	return rc;
 }
 
+static inline int
+od_console_show_errors_per_route_cb(od_route_t *route, void **argv)
+{
+	machine_msg_t *stream = argv[0];
+	assert(stream);
+
+	if (route && route->extra_logging_enabled && !od_route_is_dynamic(route)) {
```
x4m, line 342: Can we just return otherwise to reduce nesting?

x4m (changes requested):

PR #212, opened 2020-10-09
Change requied CMake >= 3.12.4

patsevanton:
Unknown CMake command "add_compile_definitions"
add_compile_definitions added in CMake 3.12.4


x4m:
Thanks!

Issue #210, opened 2020-09-17
Get odyssey configuration from CLI


bukem:
Hi guys,

Is there a way to get the configuration of the current instance of Odyssey without accessing the config file?
In Pgbouncer we have `show config;` command to get the configuration, but it seems like it is not implemented in Odyssey:
```
console=> show config;
ERROR:  odyssey: c4c171e50be88: console command error: show config;
```

Thanks!

Regards, Mikhail

x4m:
Makes sense, I think we will implement this. Or at least PR would be welcomed. Thanks for the proposal!
One caveat is that we surely will filter our credentials to mitigate security risks.

bukem:
I'd like to help with PR, but my C level is... undefined :)

x4m:
Don't worry. Just make it work. I'll help with merging. Or at least I'll find someone from YC to help you with your PR.

PR #218, opened 2020-10-20
odyssey: add command line options

reshke:
➜ ./sources/odyssey -V    
Odyssey - scalable postgresql connection poller
version 1797-358e8c4-debug
compiled with gcc version 7


reshke:
[https://github.com/yandex/odyssey/issues/210](url)

PR #224, opened 2020-10-26
fix coveirty issues + remove legacy

reshke:


docker/bin/ody-stop
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+kill $(pgrep odyssey) || (sleep 1 && kill -9 $(pgrep odyssey)) || true
```
x4m, line 2: pkill, killall, why grep?

x4m (approved):

PR #225, opened 2020-10-26
odyssey tests with asan, little test refactor

reshke:


docker/entrypoint.sh
```diff
@@ -10,4 +10,8 @@ ody-stop
 
 /ody-intergration-test
 
+ody-stop
+
+/usr/bin/odyssey-asan /etc/odyssey/odyssey.conf && /shell-test/test.sh
```
reshke, line 15: if case of some issue exit code should be != 0

reshke: ci would fail in this case

sources/frontend.c
```diff
@@ -743,17 +743,10 @@ od_frontend_remote(od_client_t *client)
 	od_server_t *server;
 	od_instance_t *instance = client->global->instance;
 	for (;;) {
-		/* we go to the eternal wait
```
x4m, line 746: is this change related to PR topic?

sources/id.h
```diff
@@ -48,7 +49,9 @@ od_id_generate(od_id_t *id, char *prefix)
 		id->id[w++] = hex[(seed[q] >> 4) & 0x0F];
 		id->id[w++] = hex[(seed[q]) & 0x0F];
 	}
+#if OD_DEVEL_LVL != -1
```
x4m, line 52: Assert if free in release build anyway

reshke: compiler warning

x4m (approved):

sources/frontend.c
reshke: this is completely wrong and should be removed,  test on this behaviour is wrong too

PR #229, opened 2020-11-20
make odyssey better

reshke:


sources/console.c
```diff
@@ -28,6 +28,7 @@ enum
 	OD_LFRONTEND,
 	OD_LROUTER,
 	OD_LVERSION,
+	OD_LHARAKIRI,
```
x4m, line 31: Do we really really want it?
Should it produce core dump or just exit?

Anyway, I think we should separate this functionality from curing memory leak which is many times more important.

x4m (changes requested):

sources/err_logger.c
```diff
@@ -74,3 +76,28 @@ od_error_logger_store_err(od_error_logger_t *l, size_t err_t)
 	od_counter_inc(l->interval_counters[l->current_interval_num], err_t);
 	return OK_RESPONSE;
 }
+
+od_retcode_t
+od_err_logger_inc_interval(od_error_logger_t *l)
+{
+	pthread_mutex_lock(&l->lock);
```
x4m, line 83: Do you really need this lock? Why not just atomic increment?

PR #250, opened 2020-12-27
change clang formatting

reshke:


x4m:
What's the reason for this?

reshke:
from https://github.com/torvalds/linux/blob/master/.clang-format

x4m (approved):
Ok, so we are going to do Linux-like formatting...

PR #263, opened 2021-01-14
 drop old server connection while online restrat in session pooling mode 

reshke:


sources/macro.h
```diff
@@ -22,4 +22,8 @@
 
 typedef int od_retcode_t;
 
+/*misc*/
+
+#define INVALID_ID -1
```
x4m, line 27: Maybe INVALID_PID?

sources/frontend.c
```diff
@@ -419,6 +419,60 @@ static inline od_frontend_status_t od_frontend_local_setup(od_client_t *client)
...
+	}
+
+	switch (client->rule->pool) {
+	case OD_RULE_POOL_SESSION: {
+		if (od_unlikely(server == NULL)) {
+			od_log(&instance->logger, "shutdown", client, server,
+			       "drop client connection on restart (session pooling)");
+			return true;
+		}
+		/*  */
```
x4m, line 446: ?

x4m (changes requested):

sources/macro.h
reshke: this is essentially either thread id orr mm coroutine id so ... just id

x4m (approved):

PR #266, opened 2021-01-22
do not hold repilation connections while online restart + drop conn rate logic

reshke:


sources/frontend.h
```diff
@@ -9,6 +9,15 @@
 
 #include "common_const.h"
 
+/* onlu for online restart */
```
x4m, line 12: onlu?

sources/frontend.h
```diff
@@ -9,6 +9,15 @@
...
+typedef struct {
+	pthread_mutex_t mu;
+	signed last_conn_drop_ts;
```
x4m, line 15: mb just int?

sources/frontend.c
```diff
@@ -419,6 +419,44 @@ static inline od_frontend_status_t od_frontend_local_setup(od_client_t *client)
 	return OD_EOOM;
 }
 
+static __thread od_drop_conn_rate_info *info = NULL;
```
x4m, line 422: so we have some thread static state of the worker elsewhere?

x4m (approved):

PR #274, opened 2021-02-09
Allow link to dynamic libpq (.so). Builds OK on ArchLinux.

rechtem:


sources/frontend.c
```diff
@@ -722,7 +724,7 @@ static od_frontend_status_t od_frontend_remote_client(od_relay_t *relay,
 		server->is_copy = 0;
 		break;
 	case KIWI_FE_QUERY:
-		if (instance->config.log_query)
+		if (instance->config.log_query && route->rule->log_query)
```
reshke, line 727: ```instance->config.log_query || route->rule->log_query```  ?

reshke:
restarted the tests - among the fallen were only flaky ones

sources/frontend.c
mrechte: Hello. I did not intend to include the second commit in this PR. However the second is to give the possibility to log queries on a per route configuration. So in order to to log queries for a given route one would have to set log_query both in the global config and in the route itself.

mrechte: May be || is better than &&

reshke: let's change here the condition of disjunction into conjunction and LGTM?

reshke (approved):
:ok: :cool:

mrechte:
What's wrong with the first commit (cee665f), it seems it has not been taken into account ?

x4m:
@mrechte everything is fine with that commit. It will be squashed during merge.

x4m:
Pushed, thanks Marc!

PR #275, opened 2021-02-11
Introduce pool client idle timeout in session pooling mode

reshke:
```current time 12868806711, drop horizon 12828797602
1155 2021-02-11T19:06:05Z info [cf80ef47cc411 s4ec75657858f] (shutdown) drop client idle connection on due timeout 20 sec
```

### DIFF
--- a/sources/frontend.c
+++ b/sources/frontend.c
@@ -477,25 +477,14 @@ static inline bool od_eject_conn_with_rate(od_client_t *client,
 static inline bool od_eject_conn_with_timeout(od_client_t *client,
 					      od_server_t *server,
 					      od_instance_t *instance,
-					      int timeout)
+					      uint64_t timeout)
 {
-	const uint64_t interval_usec = 1000000;
+	assert(server != NULL);
+	od_dbg_printf_on_dvl_lvl(1, "current time %lld, drop horizon %lld\n",
+				 machine_time_us(),
+				 client->time_last_active + timeout);
 
-	if (server == NULL) {
-		/* server is null - client was never attached to any server so its ok to eject this conn  */
-		return true;
-	}
-
-	od_dbg_printf_on_dvl_lvl(
-		1, "current time %lld, drop horizon %lld\n", machine_time_us(),
-		client->time_last_active + interval_usec * timeout);
-
-	if (client->time_last_active +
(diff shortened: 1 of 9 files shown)


sources/frontend.c
```diff
@@ -474,27 +474,77 @@ static inline bool od_eject_conn_with_rate(od_client_t *client,
...
 
-static inline bool od_should_drop_connection(od_client_t *client,
-					     od_server_t *server)
+static inline bool od_eject_conn_with_timeout(od_client_t *client,
+					      od_server_t *server,
+					      od_instance_t *instance,
+					      int timeout)
 {
-	od_instance_t *instance = client->global->instance;
+	const uint64_t interval_usec = 1000000;
```
x4m, line 482: Can we have a timeout in microseconds instead of multiplying it here and there?

sources/frontend.c
```diff
@@ -474,27 +474,77 @@ static inline bool od_eject_conn_with_rate(od_client_t *client,
...
 
-	//TODO:: drop no more than X connection per sec/min/whatever
-	if (od_likely(instance->shutdown_worker_id == INVALID_ID)) {
-		// try to optimize likely path
-		return false;
+	if (server == NULL) {
```
x4m, line 484: Are we considering any client without server expired independently from timeout?

sources/frontend.c
```diff
@@ -474,27 +474,77 @@ static inline bool od_eject_conn_with_rate(od_client_t *client,
...
+				    od_server_synchronized(server))) {
+				return od_eject_conn_with_timeout(
+					client, server, instance,
+					client->rule->pool_client_idle_timeout);
+			}
+		}
+	}
 		/* fall through */
 	case OD_RULE_POOL_TRANSACTION: {
+		//TODO:: drop no more than X connection per sec/min/whatever
```
x4m, line 536: So long-running transactions now survive pool_client_idle_timeout. Is it expected?

x4m (changes requested):

sources/frontend.c
reshke: separate setting, pool idle in transaction timeout

x4m (approved):
LGTM

PR #278, opened 2021-02-18
Check error should creation of PID file fails.

mrechte:


mrechte:
What's wrong with the failed check ?

x4m:
Hm... tests really fail each time. We'll try to investigate

PR #281, opened 2021-02-18
asynclog

reshke:


sources/logger.c
```diff
@@ -32,7 +32,52 @@ static int od_log_syslog_level[] = { LOG_INFO, LOG_ERR, LOG_DEBUG, LOG_CRIT };
...
+	(void)rc;
+}
+
+static inline void od_logger(void *arg)
+{
+	od_logger_t *logger = arg;
+
+	for (;;) {
+		machine_msg_t *msg;
+		msg = machine_channel_read(logger->task_channel, UINT32_MAX);
```
x4m, line 57: How are we going to shutdown? When Odyssey get SEGQUIT or SIGTERM?

sources/config_reader.c
```diff
@@ -122,16 +126,22 @@ static od_keyword_t od_config_keywords[] = {
 	od_keyword("log_syslog_ident", OD_LLOG_SYSLOG_IDENT),
 	od_keyword("log_syslog_facility", OD_LLOG_SYSLOG_FACILITY),
 	od_keyword("stats_interval", OD_LSTATS_INTERVAL),
+	od_keyword("asynclog", OD_LLOG_ASYNC),
```
x4m, line 129: Is there any reason to disable this?

reshke: no reason for sure

sources/logger.c
reshke: yes, just like other threads

PR #286, opened 2021-03-03
remove custom logic for sigterm

reshke:


sources/sighandler.c
```diff
@@ -100,8 +100,6 @@ void od_system_signal_handler(void *arg)
 			break;
 		switch (rc) {
 		case SIGTERM:
-			od_system_shutdown_fast(system, instance);
```
x4m, line 103: Why not remove this function?

sources/cron.c
```diff
@@ -197,15 +197,21 @@ static void od_cron_err_stat(od_cron_t *cron)
...
 	od_cron_t *cron = arg;
 	od_instance_t *instance = cron->global->instance;
 
 	cron->stat_time_us = machine_time_us();
+	pthread_mutex_lock(&cron->lock);
+	cron->online = 1;
 
 	int stats_tick = 0;
 	for (;;) {
+		if (!cron->online) {
```
x4m, line 211: In fact now we cannot shut down for 1 second. This is kind of eternity...

x4m (approved):

PR #288, opened 2021-03-04
Fix injection in auth_query

x4m:


odyssey-dev.conf
```diff
@@ -132,5 +144,5 @@ database "console" {
 locks_dir "/tmp/odyssey"
 
 graceful_die_on_errors yes
-enable_online_restart yes
+enable_online_restart no
```
reshke, line 147: why

third_party/machinarium/sources/channel.c
```diff
@@ -96,7 +96,7 @@ mm_msg_t *mm_channel_read(mm_channel_t *channel, uint32_t time_ms)
 	mm_sleeplock_lock(&channel->lock);
 
 	mm_list_t *next;
-	if (channel->msg_list_count > 0 && channel->readers_count == 0) {
+	if ((channel->msg_list_count) > 0 && (channel->readers_count == 0)) {
```
reshke, line 99: in the first case, additional parentheses are definitely not needed

reshke (approved):

third_party/machinarium/sources/channel.c
x4m: Definitely so. Fixed.

PR #296, opened 2021-03-18
Skip sys db for quantiles and use shorter window

g0djan:


sources/console.c
```diff
@@ -649,8 +649,11 @@ static inline int od_console_show_pools_add_cb(od_route_t *route, void **argv)
 					route->stats.query_hgram[i]);
 				td_merge(queries_hgram, freeze_hgram);
 			}
-			td_merge(common_transactions_hgram, transactions_hgram);
-			td_merge(common_queries_hgram, queries_hgram);
+			if (strcmp(route->id.database, "postgres") != 0 &&
+			    strcmp(route->id.database, "pgbouncer") != 0) {
```
x4m, line 653: user does not always name consiole 'pgbouncer'

sources/console.c
```diff
@@ -649,8 +649,11 @@ static inline int od_console_show_pools_add_cb(od_route_t *route, void **argv)
 					route->stats.query_hgram[i]);
 				td_merge(queries_hgram, freeze_hgram);
 			}
-			td_merge(common_transactions_hgram, transactions_hgram);
-			td_merge(common_queries_hgram, queries_hgram);
+			if (strcmp(route->id.database, "postgres") != 0 &&
```
x4m, line 652: The user may wanty 'postgres' btw

sources/console.c
g0djan: You're right. I guess we can configure a list of databases for skip via the .conf file.

PR #311, opened 2021-04-30
fallback to reuse client token to backend auth

reshke:


sources/rules.c
```diff
@@ -202,6 +202,8 @@ od_rule_t *od_rules_add(od_rules_t *rules)
 	rule->ldap_endpoint_name = NULL;
 	rule->ldap_endpoint = NULL;
 #endif
+	/* maybe some configuration here in future */
+	rule->reuse_client_passwd = 1;
```
x4m, line 206: I think this can be unconditional.

x4m (approved):

PR #313, opened 2021-05-04
Fix default_user ldap auth for default user configuration

diarworld:
If we use `user default` - ldap auth always fails, cause all searchRequets gone to LDAP server with `default_user`. Need to use `client->startup.user.value` in LDAP requests to server not `route rule name`.

Before:
Failed
After fix:
success

### DIFF
--- a/docker/ldap/odyssey.conf
+++ b/docker/ldap/odyssey.conf
@@ -54,6 +54,28 @@ database "ldap_db" {
 		quantiles "0.99,0.95,0.5"
                 client_max 107
 	}
+	user default {
+		authentication "clear_text"
+		storage "postgres_server"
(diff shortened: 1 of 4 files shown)


diarworld:
As is see ldap tests gone ok, dunno why overall tests failed =(
This feature may be useful for YC Managed Greenplum service. Cause we currently have >500 registered LDAP users in our on-prem Greenplum and it will not be easy to add all of them to config of managed service. We want to have one pool for all LDAP users and separated pool for non-LDAP tech accounts.

reshke (approved):
:cool:

reshke:
tests failed due to formatting issue. I ran the formatting, everything else looks cool

PR #315, opened 2021-05-09
Update server_login_retry on odyssey.conf

ruimarinho:
`server_login_retry` is not a listen variable, but a global one.


x4m (approved):

x4m:
Thanks! Pushed

reshke (approved):

PR #327, opened 2021-05-21
fix a couple of coverity issues

reshke:


Makefile
```diff
@@ -13,7 +13,7 @@ BUILD_TYPE=Release
 
 COMPILE_CONCURRENCY=8
 
-.PHONY: clean
+.PHONY: clean apply_fmt
```
x4m, line 17: This does not work on macOS

x4m: Ok, it seems to work...

PR #333, opened 2021-05-28
fix reload config

EvgeniiAl:
reload client_max_set variable
resolve bug when user comment 'client_max' setting, reload config and still get restriction on connections number


reshke (approved):

x4m:
Pushed, thanks!

Issue #332, opened 2021-05-28
Unable to compile on macOS `/odyssey/sources/rules.c:791:4: error: expected expression before ')' token`


ruimarinho:
```
#9 81.48 Scanning dependencies of target odyssey
#9 81.52 [  2%] Building C object sources/CMakeFiles/odyssey.dir/daemon.c.o
#9 81.69 [  3%] Building C object sources/CMakeFiles/odyssey.dir/pid.c.o
#9 81.86 [  4%] Building C object sources/CMakeFiles/odyssey.dir/logger.c.o
#9 82.15 [  5%] Building C object sources/CMakeFiles/odyssey.dir/rules.c.o
#9 82.28 /odyssey/sources/rules.c: In function 'od_rules_merge':
#9 82.28 /odyssey/sources/rules.c:578:17: warning: unused parameter 'added' [-Wunused-parameter]
#9 82.28       od_list_t *added, od_list_t *deleted)
#9 82.28       ~~~~~~~~~~~^~~~~
#9 82.28 /odyssey/sources/rules.c: In function 'od_rules_validate':
#9 82.28 /odyssey/sources/rules.c:791:4: error: expected expression before ')' token
#9 82.28     ) {
#9 82.28     ^
#9 82.29 make[2]: *** [sources/CMakeFiles/odyssey.dir/build.make:102: sources/CMakeFiles/odyssey.dir/rules.c.o] Error 1
#9 82.29 make[1]: *** [CMakeFiles/Makefile2:198: sources/CMakeFiles/odyssey.dir/all] Error 2
#9 82.29 make: *** [Makefile:130: all] Error 2
```

x4m:
Uh, really, that's a bug. Will fix soon.

ruimarinho:
Fixed by #334. Thanks @x4m.

PR #334, opened 2021-06-02
Log Parse messages when log_query enabled

x4m:


x4m:
This should also fix #332

PR #339, opened 2021-06-28
Fix parsing escaped quotes in config

RedCreepster:


x4m:
Seems like a good fix, thanks! I've made small adjustment, if you are OK with it - let's merge.

RedCreepster:
Thanks for your review. Let's merge this. Additional fixed clang formatting errors.

PR #345, opened 2021-08-14
Fix compilation without LDAP support

reshke:


x4m (approved):
Oh, so many ifdefs...

PR #346, opened 2021-08-17
Enabled writing stats in Prometheus format.

EinKrebs:


reshke (approved):

sources/logger.c
```diff
@@ -484,3 +484,49 @@ void od_logger_write(od_logger_t *logger, od_logger_level_t level,
...
+				od_route_t *route = server_ref->route;
+				is_debug = route->rule->log_debug;
+			}
+		}
+		if (!is_debug)
+			return;
+	}
+
+	int len = strlen(string);
+	char output[len + OD_LOGLINE_MAXLEN];
```
reshke, line 513: why not `char output[len + 1];` ?

sources/cron.c
```diff
@@ -238,6 +276,14 @@ void od_cron_init(od_cron_t *cron)
...
 	cron->startup_errors = 0;
 
+#ifdef PROM_FOUND
+	cron->metrics = (od_prom_metrics_t *)malloc(sizeof(od_prom_metrics_t));
+	int err = od_prom_metrics_init(cron->metrics);
+	if (err) {
+		fprintf(stdout, "Could not initialize metrics");
+	}
+#endif
+
```
reshke, line 286: maybe deallocate all memory malloc`d in od_prom_metrics_init in od_cron_stop?

reshke (commented):
in general its :ok: but we need to document this changes and new conf options in 
https://github.com/yandex/odyssey/blob/master/documentation/configuration.md
https://github.com/yandex/odyssey/blob/master/odyssey.conf

reshke (approved):

sources/logger.c
EinKrebs: Because all info in logger.format must fit in there.

PR #350, opened 2021-09-12
Update option.h

Firm:
Declare od_cli_options as static to avoid build errors.

### DIFF
--- a/sources/option.h
+++ b/sources/option.h
@@ -14,10 +14,11 @@ typedef struct {
 	int log_stdout;
 } od_arguments_t;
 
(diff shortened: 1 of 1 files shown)


Firm:
Otherwise, we could have errors like:

```
/usr/bin/ld: CMakeFiles/odyssey.dir/pid.c.o:/home/build/odyssey/sources/option.h:21: multiple definition of `od_cli_options'; CMakeFiles/odyssey.dir/daemon.c.o:/home/build/odyssey/sources/option.h:21: first defined here
```

PR #359, opened 2021-11-12
Reject extraneous data after SSL negatiation

x4m:
This prevents attacks like in CVE-2021-23214 and CVE-2021-23222.


reshke (approved):
nice

PR #366, opened 2021-11-25
fix: lint and format sh, set xtrace where necessary...

pratikbalar:
- Ignore grep in grep
- Trim spaces
- Clean cache in dpkg container image


reshke:
👍

pratikbalar:
@reshke i think tests are flaky

x4m (approved):
LGTM, thanks!

PR #369, opened 2021-11-29
Refactor tls options

reshke:
refactored code duplication around OD_CONFIG_TLS_* and OD_RULE_TLS_*


x4m:
No new CVEs expected?

reshke:
:thinking:  changes looks safe to something like (mis)configuration vulnerabilities :fearful:  etc, but we certainly need another pair of eyes :eyes:  to look on this closer

x4m (approved):

PR #371, opened 2021-12-01
Add roles in console DB

sirKuybyshev:


sources/config_reader.c
```diff
@@ -856,6 +858,30 @@ static int od_config_reader_route(od_config_reader_t *reader, char *db_name,
...
+			if (strcmp(rule->db_name, "console")) {
+				od_config_reader_error(reader, NULL, "Roles only used with console");
+				return -1;
+			}
+			rc = od_parser_next(reader->parser, &token);
+			if (rc != OD_PARSER_STRING) {
+				od_config_reader_error(
+					reader, &token,
+					"incorrect or unexpected parameter");
+				return -1;
```
reshke, line 875: `return NOT_OK_RESPONCE`

docker/odyssey.conf
```diff
@@ -3,7 +3,7 @@ storage "postgres_server" {
 	type "remote"
 	
     host "localhost"
-	port 5432
```
reshke, line 6: changes in config-examples is ok, anyway we need to provide more simple and straightforward config examples, but is that really needed?

sources/router.c
```diff
@@ -81,6 +81,27 @@ static inline int od_router_reload_cb(od_route_t *route, void **argv)
 	return 1;
 }
 
+static inline int od_drop_obsolete_rule_connections_cb(od_route_t *route,
+						       void **argv)
+{
+	od_list_t *i;
+	od_rule_t *rule = route->rule;
+	od_list_t *bad_rules = argv[0];
```
reshke, line 89: obsolete_rules

sources/console.c
```diff
@@ -1825,6 +1839,15 @@ int od_console_query(od_client_t *client, machine_msg_t *stream,
 
 	return 0;
 
+bad_role:
```
reshke, line 1842: maybe `incorrect_role`? `bad_query` is also not the best naming, we need to fix this, but not necessary in this PR

sources/rules.c
```diff
@@ -905,6 +934,17 @@ static inline char *od_rules_yes_no(int value)
 	return value ? "yes" : "no";
 }
 
+////hi, olya
+//int akkerman(int m, int n) {
```
reshke, line 938: wtf

sources/router.c
```diff
@@ -144,6 +175,13 @@ int od_router_reconfigure(od_router_t *router, od_rules_t *rules)
 			od_rule_key_free(rk);
 		}
 
+		od_list_foreach(&to_drop, i)
```
reshke, line 178: `od_list_foreach_safe` and above code should also use this MARCO, no idea why is this working without sigsegv for so long

reshke (approved):

reshke (approved):

PR #414, opened 2022-01-31
Better prometheus support

sirKuybyshev:
Fixing&extending prometheus support
Fix typo in metrics init
Add promhttp server support


config-examples/odyssey-dev.conf
```diff
@@ -50,7 +51,7 @@ listen {
 storage "postgres_server" {
 	type "remote"
 	host "localhost"
-	port 5432
+	port 5433
```
reshke, line 54: is this necessary?

reshke (approved):

Issue #432, opened 2022-04-05
Doubles rows in database statistics


clsv:
Hi, I use odyssey v1.2 in k8s as deployment in 8 replicas, in odyssey config with about 50 databases from 5 postgres13 servers. 
And I have the problem with metrics exporter, in a while exporter returns error(HTTP/1.1 500 Internal Server) with text:
```
21 error(s) occurred:
* collected metric "pgbouncer_exporter_database_current_connections" { label:<name:"name" value:"database_1" > label:<name:"pool_mode" value:"transaction" > gauge:<value:0 > } was collected before with the same name and label values
* collected metric "pgbouncer_exporter_database_current_connections" { label:<name:"name" value:"database_2" > label:<name:"pool_mode" value:"transaction" > gauge:<value:0 > } was collected before with the same name and label values
```

I have changed exporter but problem didn't disappear. After some investigation I have found cause of this problem.
When I connect to database "console" and execute query "SHOW DATABASE", I can find some rows with the same database name. I haven't understood when and how it happens. After restart everything works correctly, but in a while this problem reproduces
```
database_1          | 192.168.0.44   | 5441 | database_1          |            |       100 |            0 | transaction |               0 |                   0 |      0 |        0
database_1          | 192.168.0.44   | 5441 | database_1          |            |       100 |            0 | transaction |               0 |                   0 |      0 |        0
database_2          | 192.168.0.41   | 5442 | database_2          |            |       50 |            0 | transaction |               0 |                   0 |      0 |        0
database_2          | 192.168.0.41   | 5442 | database_2          |            |       50 |            0 | transaction |               0 |                   0 |      0 |        0
```

PR #433, opened 2022-04-21
SHOW DATABASES: fill force_user column with 'user' rule name

lazovskiy:
Currently `force_user` column is empty. This leads to an issue with some configurations, e.g.:
```
    database "foo" {
      user "user1" {
        storage "storage_foo"
        storage_user myuser
        ...
      }
      user "user2" {
        storage "storage_foo"
        storage_user myuser
        ...
      }
    }
```
In this case `SHOW DATABASES` command produces one row per database per user, but since the force_user field left empty rows are identical.
This makes the output ambiguous. It also makes hard for various exporters to deal with that data as many on them rely on `SHOW DATABASES` output.
I've chosen `user_name` field as the source for `force_user` field since it's it forms unique pair with database name whereas  `storage_user` may repeat.

The impact has been described at #432

### DIFF
--- a/sources/console.c
+++ b/sources/console.c
@@ -740,7 +740,7 @@ static inline int od_console_show_databases_add_cb(od_route_t *route,
 		goto error;
 
 	/* force_user */
-	rc = kiwi_be_write_data_row_add(stream, offset, "", 0);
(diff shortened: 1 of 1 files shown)


x4m (approved):

x4m:
Thank you! Pushed.

PR #438, opened 2022-06-21
Ldap search filter fixing

ilya-maltsev:


reshke:
Can you please describe, why current ldap search filter behavior is wrong?

ilya-maltsev:
First of all if I add to oddysey.conf line:
ldapsearchfilter "(memberOf=cn=test-db-for-ldap,cn=groups,dc=example,dc=org)"
And when I try to start oddysey I see in logs:
23707 2022-06-21T11:54:38Z info (startup) none Starting Odyssey
23707 2022-06-21T11:54:38Z error (config) none /etc/odyssey/odyssey.conf:58 unexpected parameter
23707 2022-06-21T11:54:38Z info (shutdown) none Stopping Odyssey
because in config_reader.c not included logic for parsing this param
Directly ldapsearchfilter needed for filtering LDAP-users by various conditions, for example, as mentioned above, if I want grant access only for LDAP-users, who are members of the group "test-db-for-ldap"

reshke (approved):

PR #443, opened 2022-07-01
Access management to PostgreSQL with LDAP

ilya-maltsev:
Dear maintainers @reshke @x4m !

I want to extend functionality of ldap authentication for strict access between different users to different databases on same host.
For example, we have a user "user1" who must access to database "applications" with readonly permissions (SELECT only).
And also user "user1" must access to database "proposals" with a read-write permissions (for example, SELECT, INSERT, UPDATE).
In a current way for give this permissions to database with a connection throught odyssey we need construct config like that:
```
storage "test_server" {
    host "test-server-for-ldap"
}
database "applications" {
    user "user1" {
            storage "test_server"
            storage_user "readonly-user"
            storage_password "password"            
            storage_db  "applications"
    }  
}
database "proposals" {
    user "user1" { 
            storage "test_server"
            storage_user "readwrite-user"
            storage_password "password"            
            storage_db  "proposals"
    }   
}
````
But in real environments may be too much count of users and different database on same host.
And if we going on that way - config would be not readble.

So I want to introduce a way for solve this problem with ldap authorization:
when odyssey connect to ldap for authenticate a user, after suceess authentication odyssey may receive from
ldap-server optional attributes of this user. For example it's may be a Groups.
And if we construct name of this Groups with a rules thats contains host, database and storage_user then
we can route user connection throught odyssey with a simple config, for example:

```
storage "test_server" {
    host "test-server-for-ldap"
}

ldap_endpoint "ldap1" {
}

database "default" {
    user "default" {
            storage "test_server"
            storage_user "readonly-user"
            storage_password "*******"            
            storage_db  "applications"
            ldap_endpoint_name "ldap1"

            ldap_storage_user_attr "memberof"

            ldap_storage_user "group_ro" {
                ldap_storage_username "ldap_ro"
                ldap_storage_password "password"
            }

            ldap_storage_user "group_rw" {
                ldap_storage_username "ldap_rw"
                ldap_storage_password "password"
            }
     }
}
```
For example, if we wants to grant access to "user1" with readonly permissions to database "trade_offers" -
just need to invite "user1" to Group called "test-server-for-ldap_trade_offers_group_ro" on LDAP-server.

But in current state of this PR a have **problem** with a race between clients who connects to odyssey.
The problem is that in after first step of connection to odyssey open and close connection to Backend.
And if we have two clients in a pending state:
```
console=> show clients;
 type |  user   | database  |  state  |     addr      | port  | local_addr  | local_port | con 
------+---------+-----------+---------+---------------+-------+-------------+------------+-
 C    | root    | console   | pending | 127.0.0.1     | 48640 | 127.0.0.1   |       6432 |  
 C    | user1   | test-db1  | pending | 192.168.3.242 | 50976 | 172.17.1.88 |       6432 |   
 C    | user2   | test-db2  | pending | 192.168.3.242 | 50980 | 172.17.1.88 |       6432 |       

```
then first of clients who attach to backend will be use storage_user credential of last client who setup connection to backend 
(in od_frontend_setup_params).

I will be very appreciative if you could help me solve this problem.
In this case, I see 2 solutions:
1) somehow, when a client connects to the backend, update (or recheck) in the connection variables to the backend exactly those credentials that were received through ldap authorization
(in at that way I find how set password again, but I could not change username for connect to PostgreSQL)
2) after a successful first connection to the backend, do not close the connection, but transfer the client to the active state
(at least not to authenticate a second time)

### CODE REVIEW

File: third_party/kiwi/kiwi/be_read.h [hunk 1]
```diff
@@ -17,6 +17,8 @@ struct kiwi_be_startup {
 	kiwi_var_t user;
 	kiwi_var_t database;
 	kiwi_var_t replication;
+	kiwi_var_t lsu_username;
```
reshke, line 20: Storing such settings in client->rule instead of client->startup would be better i think.

config-examples/odyssey-dev-ldap.conf
```diff
@@ -178,6 +179,19 @@ database "postgres2" {
 		ldap_endpoint_name "ldap1"
 		password_passthrough yes
 
+		#ldap_storage_user_attr "memberof"
```
reshke, line 182: ohh, `odyssey-dev*.conf` was a mistake... 
please drop this changes

sources/backend.c
```diff
@@ -106,6 +106,15 @@ static inline int od_backend_startup(od_server_t *server,
 {
 	od_instance_t *instance = server->global->instance;
 	od_route_t *route = server->route;
+
+	if (client->rule->ldap_storage_user_attr) {
```
reshke, line 110: Why do we need this changes? client route id should not change during client lifecycle, because user-server routing uses route username for server connection pooling and other logic

reshke: look for sources/router.c:334, i think we can remove this logic from here and add to router.c

reshke (changes requested):

reshke:
I will try to look for race issues, the PR in general is going in the right way

third_party/kiwi/kiwi/be_read.h
ilya-maltsev: I'm tried to store this settings in client->rule at [debug-branch](e7085dd45d59d91028965b2954fb6d1753422be2), but this variable has overridden somewhere near [od_fronted_startup](https://github.com/yandex/odyssey/blob/f9a1ec21bc908ce4814b3ca2cfc7f1028204e466/sources/frontend.c#L114) by another client and race come again...

ilya-maltsev:
The race problem between different clients was solved, but I got another problem:
for example, the first time a client connects to a backend with an LDAP group that gave it ldap_storage_user as ldap_dba.
Then, if we change the group for this client and the LDAP group gives it ldap_storage_user as ldap_rw, then the next time the client wants to connect, it will connect with the previous session's authorization credentials:

210510 2022-07-22T07:06:20Z info startup new client connection
210510 2022-07-22T07:06:20Z debug tls ssl request
210510 2022-07-22T07:06:20Z debug tls is disabled, ignoring
210510 2022-07-22T07:06:20Z debug [none-none] [none-none] routing matched rule: default_db default_user with (null) routing type
210510 2022-07-22T07:06:20Z info startup route 'cryptomax.odyssey' to 'default_db.default_user'
210510 2022-07-22T07:06:20Z debug auth PasswordMessage
210510 2022-07-22T07:06:20Z debug auth checking passwd against ldap endpoint ldap1
210510 2022-07-22T07:06:20Z debug auth_ldap 0 connections are currently issued to ldap
210510 2022-07-22T07:06:20Z debug auth_ldap_debug matched group CN=test1-db-ldap_database-1_group_rw,CN=Users,DC=example,DC=local
210510 2022-07-22T07:06:20Z debug auth_ldap new ldap_storage_user: ldap_rw
210510 2022-07-22T07:06:20Z debug setup sending params:
210510 2022-07-22T07:06:20Z debug setup TimeZone = Europe/Moscow
210510 2022-07-22T07:06:20Z debug setup standard_conforming_strings = on
210510 2022-07-22T07:06:20Z debug setup session_authorization = ldap_dba
210510 2022-07-22T07:06:20Z debug setup server_version = 11.5
210510 2022-07-22T07:06:20Z debug setup server_encoding = UTF8
210510 2022-07-22T07:06:20Z debug setup is_superuser = off
210510 2022-07-22T07:06:20Z debug setup IntervalStyle = postgres
210510 2022-07-22T07:06:20Z debug setup integer_datetimes = on
210510 2022-07-22T07:06:20Z debug setup DateStyle = ISO, DMY
210510 2022-07-22T07:06:20Z debug setup client_encoding = UTF8
210510 2022-07-22T07:06:20Z debug setup application_name = psql

I think the solution to this problem is need to look at storage_user when attach client to server...

sources/router.c
```diff
@@ -306,7 +306,9 @@ od_router_status_t od_router_route(od_router_t *router, od_client_t *client)
 {
 	kiwi_be_startup_t *startup = &client->startup;
 	od_instance_t *instance = router->global->instance;
-
+#ifdef LDAP_FOUND
+	od_ldap_server_t *ldap_server = od_ldap_server_allocate();
```
ilya-maltsev, line 310: When I try to store a pointer to ldap_server in client->ldap_server or rule->ldap_server, I get an error:
/odyssey/sources/client.h:73:9: error: unknown type name ‘od_ldap_server_t’
   73 |         od_ldap_server_t *ldap_server;
How fix it?

reshke:
Fow now code changes looks ok. Can we add some tests to `docker/ldap/test ldap.sh` before merge? Also docs changes will be very appreciated

ilya-maltsev:
ok, I will try to add tests and docs

ilya-maltsev:
> HI Ilya, hope you are doing well and safe. So, is this pr ready for review? If it this, please click the "ready for review" button.

Hello @reshke , thank you for your concern)
In this PR state, the main logic for selecting ldap_storage_credential does not use ldap_pool because ldap_pool stores on a route that is not defined when it is necessary to select ldap_server from the poll. So we need to open a new connection to the ldap server, bind, search, and close the connection each time the client logs in. Now I'm trying to move ldap_pool to another place, but since ldap_pool uses the same structures as the pool for pg servers, it's not easy... But I think I've found a way, and soon the PR will be ready for review.

ilya-maltsev:
All commits are merged into one in a new branch and are ready to review in [Access management to PostgreSQL with LDAP](https://github.com/yandex/odyssey/pull/454)

PR #446, opened 2022-07-24
Add aarch64 support for machinarium context switch function

reshke:
Implement mm_context_swap function for AArch64 (64-bit ARM) architecture.
The purpose of mm_context_swap is to save all caller-safe registers to stack, then save stack pointer to arg0, and  change stack pointer to arg1. After stack pointer is changed, registers are popped back, `restoring` new running context.
Then we need to return somewhere, executing `ret` asm cmd. The return address should point to new context running function. 

That is, we enter mm_context_swap function in one coroutine and exit in other (not the same). 

On x86_64 or i386 architecture this is enforced by placing return address directly on stack  in context.c:37. The newly allocated stack will pop zero values to its registers and then execute ret cmd, returing to adrees is current stack pointer contains. 

On AArch64 `ret` command works in a different way: it returns to address stored in some register, x30 for default. This is why `mm_context_prepare` behaves differently in this case: we place return address to `*(sp + 1) `, this is exactly address which will be in x30 registers after context creation and block of `ldp` commands

The 64-bit ARM (AArch64) calling convention allocates the 31 general-purpose registers.

x31 (SP) is a stack pointer, we dont need to save in during `mm_context_swap_function`. 

Saved during `mm_context_prepare`:
x30 is procedure link register, x29 frame pointer, we need to store them, also we need to save x19 to x29, which is callee-saved registers. 
x18 is platform register, x16 and x17 intra-procedure-call scratch registers, i decided its better to save them then not.
x8 indirect return value address, i thick its better to save such thing.

Not saved during `mm_context_prepare`:
x9 to x15: Local variables, caller saved, so dont need to save on stack.
x0 to x7: Argument values passed to and results returned from a subroutine, we dont save them.

TODO: 
Add docker images wich odyssey test on Aarch64, also provide some deb-packages

### DIFF
--- a/third_party/machinarium/sources/context.c
+++ b/third_party/machinarium/sources/context.c
@@ -33,12 +33,20 @@ static inline void **mm_context_prepare(mm_contextstack_t *stack)
 {
 	void **sp;
 	sp = (void **)(stack->pointer + stack->size);
+#if __amd64
 	*--sp = NULL;
 	*--sp = (void *)mm_context_runner;
-#if __amd64
+	/* for x86_64 we need to place return address on stack */
 	sp -= 6;
 	memset(sp, 0, sizeof(void *) * 6);
+#elif __aarch64__
+	/* for aarch64 we need to place return address in x30 reg */
+	sp -= 16;
+	memset(sp, 0, sizeof(void *) * 16);
+	*(sp + 1) = (void *)mm_context_runner;
 #else
+	*--sp = NULL;
+	*--sp = (void *)mm_context_runner;
 	sp -= 4;
 	memset(sp, 0, sizeof(void *) * 4);
 #endif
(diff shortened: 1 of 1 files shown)


reshke:
just a nice [pic](https://www.processon.com/view/link/6295e1b80791294815547608)

third_party/machinarium/sources/context.c
```diff
@@ -33,12 +33,20 @@ static inline void **mm_context_prepare(mm_contextstack_t *stack)
...
 	sp = (void **)(stack->pointer + stack->size);
+#if __amd64
 	*--sp = NULL;
 	*--sp = (void *)mm_context_runner;
-#if __amd64
+	/* for x86_64 we need to place return address on stack */
 	sp -= 6;
 	memset(sp, 0, sizeof(void *) * 6);
+#elif __aarch64__
+	/* for x86_64 we need to place return address in x30 reg */
```
x4m, line 43: typo in arch name

third_party/machinarium/sources/context.c
```diff
@@ -64,41 +72,71 @@ void mm_context_create(mm_context_t *context, mm_contextstack_t *stack,
 	mm_context_swap(&context_runner, context);
 }
 
-#if !defined(__amd64) && !defined(__i386)
+#if !defined(__aarch64__) && !defined(__amd64) && !defined(__i386)
 #error unsupported architecture
 #endif
 
-asm("\t.text\n"
-    "\t.globl mm_context_swap\n"
```
x4m, line 72: First two asm lines are still common. Did you copied them intentionally so that for each arch we have valid ASM chunk?

x4m (changes requested):

third_party/machinarium/sources/context.c
reshke: yes

x4m (approved):

PR #455, opened 2022-08-16
Nits: include headers for explicit function declaration, drop dead code

reshke:
Just a bunch of nits


x4m (approved):
How about dithcing mutex in favor of od spinlock?

PR #456, opened 2022-08-16
Target session attr feature for storage

reshke:
Storage host section is now accepts a comma-separated
PostgreSQL hosts list, in format "host2,host2"
or "[host1]:port1,host2" etc.

Storage structure now holds pointer to host, which connection was acquired.
This is required to cancel`s queries to work fine.

Target session attrs policies added:

* read-write
* read-only
* any (default)

Target session attrs uses pg_is_in_recovery function to
check if host in primary or replica. in_hot_standby option is not checked yet.

### DIFF
--- a/config-examples/odyssey-dev.conf
+++ b/config-examples/odyssey-dev.conf
@@ -50,8 +50,9 @@ listen {
 
 storage "postgres_server" {
 	type "remote"
-	host "localhost"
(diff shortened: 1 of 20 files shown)


usernamedt (approved):
We can make the `[]` non-mandatory to allow the `host1:port1,host2` format for non-ipv6 hostnames. But I guess it can be added in future versions.

PR #457, opened 2022-08-22
Fix undefined behaviour in od_worker_pool_feed function

reshke:
Currently od_worker_pool_feed function uses non-atomic access to
pool->round_robin variable. But od_worker_pool_feed function can be
called multiple time simultaneously from od_system_server, as odyssey
supports multiple listen addresses feauture, and no synchronization is
performed between concurrent listen coroutines. This entails data race, which
leads undefined behaviour in C/C++

### DIFF
--- a/sources/worker_pool.h
+++ b/sources/worker_pool.h
@@ -13,8 +13,8 @@ typedef struct od_worker_pool od_worker_pool_t;
 
 struct od_worker_pool {
 	od_worker_t *pool;
-	int round_robin;
-	int count;
+	od_atomic_u32_t round_robin;
+	uint32_t count;
 };
 
 static inline void od_worker_pool_init(od_worker_pool_t *pool)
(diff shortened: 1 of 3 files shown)


sources/worker_pool.h
```diff
@@ -74,12 +75,17 @@ od_worker_pool_wait_gracefully_shutdown(od_worker_pool_t *pool)
...
-	int next = pool->round_robin;
-	if (pool->round_robin >= pool->count) {
-		pool->round_robin = 0;
-		next = 0;
+	uint32_t next;
+	uint32_t oldValue;
+
+	while (1) {
+		oldValue = od_atomic_u32_of(&pool->round_robin);
+		next = oldValue + 1 == pool->count ? 0 : oldValue;
```
x4m, line 83: just use od_atomic_u64_inc()

x4m (approved):

PR #469, opened 2022-09-16
ldap_pool_ttl

ilya-maltsev:
To avoid using a broken ldap connection when getting it from the pool, need to add the ttl parameter to ldap_server.


reshke (approved):
Thanks, this is very usefull setting.

Issue #478, opened 2022-12-13
Odysseus crashes without issue description after connection LDAP user


skilyazhnev:
Hi!

We faced a problem when we build odyssey with LDAP lib (we build in Odessey docker-dev env)

We must increase `coroutine_stack_size` from 4 to 16 or odyssey fall with a `segment fault` error after an attempt to connect any LDAP user. 

And in odyssey-dev-ldap.conf you already did this. I guess Odyssey need some error notification or doc note about running odyssey with LDAP. 

Because we were only able to find Odyssey problem through tests and this can be tricky for new users using LDAP.

x4m:
Hi! Yes, you are absolutely correct. LDAP routines needs more stack than typically Odyssey allocates. Please, send a PR with such a note that would simplify life for anyone following your route.
Thank you for letting us know!

PR #480, opened 2023-01-11
Added config check for coroutine_stack_size

ilya-maltsev:


ilya-maltsev:
Looks like it's time to renew the [certificates](https://github.com/yandex/odyssey/blob/76ea5f7169b26a70a4bf20cb990ea30ab8691d52/test/machinarium/server.crt#L1)

pmwkaa:
I think we should investigate the idea of using `mmap(MAP_GROWSDOWN)` for stack instead of relying on the fixed stack size. I don't have much time to take a look, but I think it is doable

ilya-maltsev:
simply adding the MAP_GROWSDOWN flag into [mmap](https://github.com/yandex/odyssey/blob/76ea5f7169b26a70a4bf20cb990ea30ab8691d52/third_party/machinarium/sources/context_stack.c#L19) doesn't help

pmwkaa:
This is not how it works, you should find a working example how to implement it, it is more complex

ilya-maltsev:
Some people say: "Trying to rely on the MAP_GROWSDOWN flag is unreliable and dangerous because it cannot protect you against mmap creating a new mapping just adjacent to your stack, which will then get clobbered."
Also in many of last commits on github people removing using of MAP_GROWSDOWN.
May be I don't undestend concept, but for me it is not best idea for finding optimal stack size.
I think it's must be more simply and careful.

pmwkaa:
The idea there is that you allow to grow stack automatically up to some adequate limit. It is better to find an implementation, how people are using it in their projects. It is not necessary about the GROWSDOWN flag, we need to allocate memory for stack usage. The direction of stack grows itself might be different depending on the architecture

ilya-maltsev:
Great idea, but I think dynamic memory allocation is not in best practices for high load projects, anyway for C-based, for example in nginx if you create many virtual servers more than placed in memory you get an error:
`Restarting nginx: nginx: [emerg] could not build the server_names_hash, 
you should increase either server_names_hash_max_size: 256 
or server_names_hash_bucket_size: 64.` 
although it would seem, why not use a predetermined limit for some simple name table hash?
but seems not worth the trouble

ilya-maltsev:
@pmwkaa 
is it possible to replace mmap with malloc at [mm_contextstack_create](https://github.com/yandex/odyssey/blob/76ea5f7169b26a70a4bf20cb990ea30ab8691d52/third_party/machinarium/sources/context_stack.c#L19)  ?
in my local_build this approach solved the problem...

ilya-maltsev:
Problem does not in dynamic resizing, its was described in [issue](https://github.com/yandex/odyssey/issues/478)
And also when PAM auth was used (as [reshke](https://github.com/reshke) [said](https://github.com/yandex/odyssey/pull/480#issuecomment-1413760842)).
As a result, if use malloc instead of mmap odyssey does not crash with serfault when using too small values of the coroutine_stack_size

ilya-maltsev:
you goddamn right)

ilya-maltsev:
@pmwkaa 
As a result for odyssey correct work with ldap_auth coroutine_stack_size must be more 8 and for pam_auth  coroutine_stack_size must be more 128 
Sorry but I can't see a way to solve this problem with dynamic memory allocation, because the MAP_GROWSDOWN flag of mmap allows the stack to grow downward into the new page, but it does not automatically allocate the new page and using malloc instead of mmap doesn't solve the problem correctly either.

reshke:
> Looks like it's time to renew the [certificates](https://github.com/yandex/odyssey/blob/76ea5f7169b26a70a4bf20cb990ea30ab8691d52/test/machinarium/server.crt#L1)

can you create separate PR for this change?

ilya-maltsev:
> > Looks like it's time to renew the [certificates](https://github.com/yandex/odyssey/blob/76ea5f7169b26a70a4bf20cb990ea30ab8691d52/test/machinarium/server.crt#L1)
> 
> can you create separate PR for this change?

https://github.com/yandex/odyssey/pull/492

rkhapov:
seemd like this pr is abandoned

PR #488, opened 2023-04-14
Add pool_discard_query option

pensnarik:
In some PostgreSQL systems user permissions are not granted explicitly, they use `ROLE` session variable instead. For instance: `ALTER USER <login_user> SET ROLE <permission role>`. When Odyssey is in "smart_discard" mode and returns a connection to the pool, several statements are executed, including `SET SESSION AUTHORIZATION DEFAULT`. This causes the session `ROLE` variable to reset to "none", resulting in the loss of its initial value:
```
2023-04-14 14:19:40 UTC [1326951] => [501-1] client=127.0.0.1,db=<db>,user=<user> LOG:  duration: 0.160 ms  statement: SET SESSION AUTHORIZATION DEFAULT;RESET ALL;CLOSE ALL;UNLISTEN *;SELECT pg_advisory_unlock_all();DISCARD PLANS;DISCARD SEQUENCES;DISCARD TEMP;
```
I think Odyssey developers used all statements that execute when `DISCARD ALL` happens and removed `DEALLOCATE ALL` from them and thus `SET SESSION AUTHORIZATION DEFAULT` is also executed. After this statement is executed the session `ROLE` variable resets to `none` and its initial values is lost.
The new settings `pool_no_reset_auth` allows to skip `SET SESSION AUTHORIZATION DEFAULT` statement in smart discard mode.
```
2023-04-14 14:18:48 UTC [1326725] => [496-1] client=127.0.0.1,db=<db>,user=<db> LOG:  duration: 0.103 ms  statement: RESET ALL;CLOSE ALL;UNLISTEN *;SELECT pg_advisory_unlock_all();DISCARD PLANS;DISCARD SEQUENCES;DISCARD TEMP;
```
This setting should be used in tandem with `smart_discard`. When both `smart_discard` and `pool_no_reset_auth` are enabled `SET SESSION AUTHORIZATION DEFAULT` will not be executed when Odyssey returns connection to the pool.

### DIFF
--- a/documentation/configuration.md
+++ b/documentation/configuration.md
@@ -685,6 +685,27 @@ from the pool.
 
 `pool_discard no`
 
+#### pool\_smart\_discard *yes|no*
+
+When this parameter is enabled, Odyssey sends smart discard query instead of default `DISCARD ALL` when it
+returns connection to the pool. Its default value may be overwritten by pool\_discard\_string setting.
+
+#### pool\_discard\_string *string*
+
+When resetting a database connection, a pre-defined query string is sent to the server. This query string consists of a set of SQL statements that will be executed during a `DISCARD ALL` command, except for `DEALLOCATE ALL`. The default query string includes the following statements:
+
+```sql
+SET SESSION AUTHORIZATION DEFAULT;
+RESET ALL;
+CLOSE ALL;
+UNLISTEN *;
+SELECT pg_advisory_unlock_all();
+DISCARD PLANS;
+DISCARD SEQUENCES;DISCARD TEMP;
+```
+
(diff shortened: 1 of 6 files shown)


pensnarik:
After discussion with @x4m we decided to implement a new settings called `pool_discard_string` which will allow to set a custom discard string.

x4m:
Thank you for your contribution! This PR looks good to me, but I have some nits.
1. I think **pool_discard_string** should be activated independently of **smart_discard** setting.
2. It would be nice to warn user about having deallocate in **pool_discard_string** when reserved prepared statments are configured.

x4m (approved):

PR #497, opened 2023-05-22
sources/hashmap.c: partially handle missing malloc failure check

chipitsine:
there're two missing checks, second will go as separate PR


reshke (approved):

reshke (approved):

reshke:
Thanks!

PR #504, opened 2023-06-06
CI: remove dedicated "clang-format" workflow in flavour of "cmake"

chipitsine:
"cmake" workflow already performs format check. also, "clang-format" is broken


reshke (approved):

reshke:
thanks!

PR #510, opened 2023-07-07
Fix ch1nn1y odyssey null logger 0

AndrewOvvv:
write if statement for checking is logger value equals to NULL

### DIFF
--- a/sources/module.c
+++ b/sources/module.c
@@ -46,10 +46,14 @@ int od_target_module_add(od_logger_t *logger, od_module_t *modules,
 
 	handle = od_dlopen(target_module_path);
 	if (!handle) {
-		od_log(logger, "load_module", NULL, NULL,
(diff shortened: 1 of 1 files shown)


sources/module.c
```diff
@@ -46,10 +46,14 @@ int od_target_module_add(od_logger_t *logger, od_module_t *modules,
...
-		od_log(logger, "load_module", NULL, NULL,
-		       "errors while openning %s module: %s",
-		       target_module_path, dlerror());
-		goto error;
+            if (logger) {
+                od_log(logger, "load_module", NULL, NULL,
+                        "errors while openning %s module: %s",
+                        target_module_path, dlerror());
+            } else {
+                printf("error: logger does not exist (address setted to NULL value)\n");
```
reshke, line 54: Error msg should be exactly as above

reshke (approved):

PR #513, opened 2023-07-14
Fix for segfault on prometheus metrics logging

ilya-maltsev:
Based on #508

### DIFF
--- a/sources/cron.c
+++ b/sources/cron.c
@@ -71,8 +71,8 @@ static int od_cron_stat_cb(od_route_t *route, od_stat_t *current,
 		if (instance->config.log_route_stats_prom) {
 			const char *prom_log =
 				od_prom_metrics_get_stat_cb(metrics);
-			od_logger_write_plain(&instance->logger, OD_LOG,
-					      "stats", NULL, NULL, prom_log);
+			od_log(&instance->logger, "stats", NULL, NULL,
+			       prom_log);
 			free(prom_log);
 		}
 	}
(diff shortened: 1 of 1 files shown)


ilya-maltsev:
changes only in sources/cron.c....
sources/frontend.c - realy need to be formatted?

reshke:
> changes only in sources/cron.c.... sources/frontend.c - realy need to be formatted?
Yes, seems that.

If this is true resolves issue, we can actually drop function od_logger_write_plain completely

ilya-maltsev:
after this changes prometheus metrics in syslog looks like:
```
Jul 14 06:02:49 test-odyssey odyssey: 1716060 2023-07-14T06:02:49Z info [none-none] [none-none] stats # HELP client_pool_total Total database clients count\n# TYPE client_pool_total gauge\nclient_pool_total{database\="default_db"} 1\n\n# HELP avg_tx_count Average transactions count per second\n# TYPE avg_tx_count gauge\navg_tx_count{user\="default_user",database\="default_db"} 0\n\n# HELP avg_tx_time Average transaction time in usec\n# TYPE avg_tx_time gauge\navg_tx_time{user\="default_user",database\="default_db"} 0\n\n# HELP avg_query_count Average query count per second\n# TYPE avg_query_count gauge\navg_query_count{user\=
```
but size of prom_metrics more then OD_LOGLINE_MAXLEN https://github.com/yandex/odyssey/blob/de02985941c0a375f9373aa78efd5abed56d44f7/sources/logger.h#L10

and with od_log all of prometheus metrics cannot be placed in logfile

so I would like to suggest to remove this functionality which was added in earlier implementations of prometheus metrics support - #346

rkhapov:
seems like this pr is abandoned

PR #531, opened 2023-09-01
check for replication lag before every statement dispatch

reshke:


x4m:
This might hang amidst message body.

rkhapov:
already implemented

PR #537, opened 2023-10-31
Added config testing (odyssey --test)

NikitaUnisikhin:


docker/config-validation/configs/workers/valid/odyssey-test3.conf
```diff
@@ -0,0 +1,7 @@
+workers "auto"
+
+log_format "%p %t %l [%i %s] (%c) %m\n"
+
+listen {
```
reshke, line 5: maybe configs without any rules should not be valid, as they are not usable?

NikitaUnisikhin: done

config-examples/odyssey-aq.conf
```diff
@@ -117,6 +117,7 @@ database "postgres" {
 		auth_query_db "postgres"
 		storage_password "passwd"
 
+		storage "postgres_server"
```
reshke, line 120: ?

.DS_Store
reshke, line 1: drop it

NikitaUnisikhin: done

reshke, line 1: this binary still present
also, maybe modify .gitignore file to always omit this on commit?

NikitaUnisikhin: done

config-examples/odyssey-aq.conf
NikitaUnisikhin: fixed

reshke (approved):

PR #550, opened 2023-12-11
Add addr pool in rule

NikitaUnisikhin:
-- Added address to the rule. Address supported IPv4/IPv6 with mask and hostname
-- Сhanged the logs taking into account the new parameter
-- Сreated a module for working with addresses (address.c) also used the module in hba
-- Added tests for new functionality

### DIFF
--- a/sources/address.c
+++ b/sources/address.c
@@ -86,6 +86,20 @@ int od_address_read(struct sockaddr_storage *dest, const char *addr)
 	return -1;
 }
 
+static bool od_address_ipv4eq(struct sockaddr_in *a, struct sockaddr_in *b)
+{
+	return (a->sin_addr.s_addr == b->sin_addr.s_addr);
+}
+
+static bool d_address_ipv6eq(struct sockaddr_in6 *a, struct sockaddr_in6 *b)
+{
+	int i;
+	for (i = 0; i < 16; i++)
+		if (a->sin6_addr.s6_addr[i] != b->sin6_addr.s6_addr[i])
+			return false;
+	return true;
+}
+
 bool od_address_equals(struct sockaddr_storage *firstAddress,
 		       struct sockaddr_storage *secondAddress)
 {
@@ -93,18 +107,11 @@ bool od_address_equals(struct sockaddr_storage *firstAddress,
 		return false;
 
 	if (firstAddress->ss_family == AF_INET) {
-		struct sockaddr_in *addr1 = (struct sockaddr_in *)firstAddress;
-		struct sockaddr_in *addr2 = (struct sockaddr_in *)secondAddress;
-		return (addr1->sin_addr.s_addr ^ addr2->sin_addr.s_addr) == 0;
+		return od_address_ipv4eq((struct sockaddr_in *)firstAddress,
+					 (struct sockaddr_in *)secondAddress);
 	} else if (firstAddress->ss_family == AF_INET6) {
-		struct sockaddr_in6 *addr1 = (struct sockaddr_in6 *)firstAddress;
-		struct sockaddr_in6 *addr2 = (struct sockaddr_in6 *)secondAddress;
-		for (int i = 0; i < 16; ++i) {
-			if (addr1->sin6_addr.s6_addr[i] ^ addr2->sin6_addr.s6_addr[i]) {
-				return false;
-			}
-		}
-		return true;
+		return d_address_ipv6eq((struct sockaddr_in6 *)firstAddress,
+					(struct sockaddr_in6 *)secondAddress);
 	} else if (firstAddress->ss_family == AF_UNSPEC) {
 		return true;
 	}
@@ -242,7 +249,8 @@ bool od_address_check_hostname(struct sockaddr_storage *client_sa, const char *h
 	found = false;
 	for (gai = gai_result; gai; gai = gai->ai_next)
 	{
-		if (od_address_equals((struct sockaddr_storage *)gai->ai_addr, client_sa))
+		if (od_address_equals((struct sockaddr_storage *)&gai->ai_addr),
+		                      (struct sockaddr_storage *)client_sa)
 		{
 			found = true;
 			break;
--- a/sources/address.c
+++ b/sources/address.c
@@ -167,7 +167,7 @@ uint32 od_address_bswap32(uint32 x)
 	       ((x >> 8) & 0x0000ff00) | ((x >> 24) & 0x000000ff);
 }
 
-bool od_address_hostname_match(const char *pattern, const char *actual_hostname)
+static bool hostname_match(const char *pattern, const char *actual_hostname)
 {
 	if (pattern[0] == '.')		/* suffix match */
 	{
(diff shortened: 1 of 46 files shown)


sources/address.c
```diff
@@ -0,0 +1,281 @@
...
+			if (ch2 >= 'A' && ch2 <= 'Z')
+				ch2 += 'a' - 'A';
+			else if (IS_HIGHBIT_SET(ch2) && isupper(ch2))
+				ch2 = tolower(ch2);
+
+			if (ch1 != ch2)
+				return (int) ch1 - (int) ch2;
+		}
+		if (ch1 == 0)
+			break;
```
AndrewOvvv, line 152: Почему мы считаем, что s1 короче s2? (типа, почему мы только ch1 проверяем)

sources/address.c
```diff
@@ -0,0 +1,281 @@
...
+
+	/* Lookup IP from host name and check against original IP */
+	ret = getaddrinfo(client_hostname, NULL, NULL, &gai_result);
+	if (ret != 0)
+		return false;
+
+	found = false;
+	for (gai = gai_result; gai; gai = gai->ai_next)
+	{
+		if (gai->ai_addr->sa_family == client_sa->ss_family)
```
AndrewOvvv, line 202: Не суверен, но разве od_address_equals не делает всё то, что  происходит внутри?

sources/address.c
```diff
@@ -0,0 +1,281 @@
...
+	char *end = NULL;
+	long len = strtoul(prefix, &end, 10);
+	if (*prefix == '\0' || *end != '\0') {
+		return -1;
+	}
+	if (address_range->addr.ss_family == AF_INET) {
+		if (len > 32)
+			return -1;
+		struct sockaddr_in *addr = (struct sockaddr_in *)&address_range->mask;
+		long mask;
```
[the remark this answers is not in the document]
AndrewOvvv: I think that it is not good idea to used signed type for mask. There're exist some UB moments with signed integer types and bitwise operations

sources/address.c
```diff
@@ -0,0 +1,281 @@
...
+		return true;
+	}
+
+	return false;
+}
+
+bool od_address_range_equals(od_address_range_t *first, od_address_range_t *second)
+{
+	if (first->is_hostname == second->is_hostname)
+		return first->string_value == second->string_value;
```
x4m, line 123: od_address_strcasecmp()?

sources/address.c
x4m: It's expected that ch1==ch2 at this point...
but, idk, there's strcasecmp() https://pubs.opengroup.org/onlinepubs/9699919799/functions/strcasecmp.html

sources/address.c
```diff
@@ -0,0 +1,281 @@
...
+	if (reti)
+		return -1;
+	reti = regexec(&regex, hostname, 0, NULL, 0);
+	if (reti == 0)
+		return 0;
+	else
+		return 1;
+}
+
+uint32 od_address_bswap32(uint32 x)
```
x4m, line 277: BTW we have od_hba_bswap32().
But I'd prefer to have od_bswap32() once and for all :)

sources/router.c
```diff
@@ -348,14 +349,24 @@ od_router_status_t od_router_route(od_router_t *router, od_client_t *client)
 
 	/* match latest version of route rule */
 	od_rule_t *rule;
+
+	struct sockaddr_storage sa;
+	if (!client->is_watchdog) {
```
x4m, line 354: Why do we exempt watchdog? Can we exampt all outgoing clients, not just watchdog?

sources/address.c
```diff
@@ -0,0 +1,281 @@
...
+{
+	od_address_range_t address_range = {
+		.string_value = strdup("all"),
+		.string_value_len = strlen("all"),
+		.is_default = 1
+	};
+	return address_range;
+}
+
+int od_address_range_copy(od_address_range_t *src, od_address_range_t *dst)
```
x4m, line 21: maybe free() hostname?

sources/rules.c
```diff
@@ -178,6 +178,8 @@ void od_rules_rule_free(od_rule_t *rule)
 		free(rule->db_name);
 	if (rule->user_name)
 		free(rule->user_name);
+	if (rule->address_range.string_value)
```
x4m, line 181: maybe free address_range too?

x4m (changes requested):
Generally looks good to me, however feels very leaky. Let's make sure all allocations are free()ed correctly.

sources/address.c
NikitaUnisikhin: this case is considered in the next iteration of the loop

sources/address.c
```diff
@@ -0,0 +1,281 @@
...
+	dst->addr = src->addr;
+	dst->mask = src->mask;
+	dst->is_default = src->is_default;
+	dst->is_hostname = src->is_hostname;
+}
+
+int od_address_range_read_prefix(od_address_range_t *address_range, char *prefix)
+{
+	char *end = NULL;
+	long len = strtoul(prefix, &end, 10);
```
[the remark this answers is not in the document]
NikitaUnisikhin: length can be <= zero (str 55)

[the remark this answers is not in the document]
AndrewOvvv: but why do you use strtoul (to unsigned long)?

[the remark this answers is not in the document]
NikitaUnisikhin: oh yeah, thanks for the comment, I'll fix it

sources/address.c
```diff
@@ -28,28 +26,31 @@ int od_address_range_copy(od_address_range_t *src, od_address_range_t *dst)
...
 	if (address_range->addr.ss_family == AF_INET) {
 		if (len > 32)
 			return -1;
-		struct sockaddr_in *addr = (struct sockaddr_in *)&address_range->mask;
-		long mask;
+		struct sockaddr_in *addr =
+			(struct sockaddr_in *)&address_range->mask;
+		uint32 mask;
 		if (len > 0)
 			mask = (0xffffffffUL << (32 - (int)len)) & 0xffffffffUL;
```
AndrewOvvv, line 44: it seems that you now don't need to do "& 0xffffffffUL" here

sources/address.c
NikitaUnisikhin: thanks, fixed

sources/address.c
NikitaUnisikhin: thanks, fixed

sources/address.c
NikitaUnisikhin: thanks, done

sources/router.c
NikitaUnisikhin: thanks, now we exempt all internal clients

sources/address.c
NikitaUnisikhin: thanks, fixed

sources/rules.c
NikitaUnisikhin: I don't use malloc

sources/address.c
NikitaUnisikhin: I don't use malloc

x4m (approved):

PR #556, opened 2024-01-19
Fix parse complete msgs order in pipeline mode

reshke:
to check 
these test https://github.com/pg-sharding/spqr/tree/master/test/xproto
are working


x4m (approved):
It would bt good to have a test for this in our repo...

rkhapov:
relay arch changed, pr is outdated

PR #559, opened 2024-01-23
Add groups

NikitaUnisikhin:
-- added group authentication


sources/group.c
```diff
@@ -0,0 +1,83 @@
...
+	uint32_t size;
+	int rc;
+	rc = kiwi_read32(&size, &pos, &pos_size);
+	if (kiwi_unlikely(rc == -1))
+		goto error;
+	/* count */
+	uint16_t count;
+	rc = kiwi_read16(&count, &pos, &pos_size);
+
+	if (kiwi_unlikely(rc == -1))
```
AndrewOvvv, line 59: may be kiwi_likely here?

or if "kiwi_likely" does not exist write rc != -1

sources/group.c
```diff
@@ -0,0 +1,83 @@
...
+		goto error;
+
+	if (count != 1)
+		goto error;
+
+	/* (not used) */
+	uint32_t val_len;
+	rc = kiwi_read32(&val_len, &pos, &pos_size);
+	if (kiwi_unlikely(rc == -1)) {
```
AndrewOvvv, line 68: same as at line 59

sources/group.c
```diff
@@ -0,0 +1,83 @@
...
+
+	/* size */
+	uint32_t size;
+	int rc;
+	rc = kiwi_read32(&size, &pos, &pos_size);
+	if (kiwi_unlikely(rc == -1))
+		goto error;
+	/* count */
+	uint16_t count;
+	rc = kiwi_read16(&count, &pos, &pos_size);
```
AndrewOvvv, line 57: are you sure that you should read from pos? you have already read 4 bytes of "size" variable from here?

sources/group.c
```diff
@@ -0,0 +1,83 @@
...
+
+	if (kiwi_unlikely(rc == -1))
+		goto error;
+
+	if (count != 1)
+		goto error;
+
+	/* (not used) */
+	uint32_t val_len;
+	rc = kiwi_read32(&val_len, &pos, &pos_size);
```
AndrewOvvv, line 67: val_len is the same to size because you read it from same position

sources/config_reader.c
```diff
@@ -1203,7 +1209,8 @@ od_config_reader_ldap_storage_credentials(od_config_reader_t *reader,
 static int od_config_reader_rule_settings(od_config_reader_t *reader,
 					  od_rule_t *rule,
 					  od_extention_t *extentions,
-					  od_storage_watchdog_t *watchdog)
+					  od_storage_watchdog_t *watchdog,
+					  od_group_t *group)
```
AndrewOvvv, line 1213: why you gives group as argument? you have already had it in "od_rule_t *rule"

sources/config_reader.c
```diff
@@ -1724,7 +1744,79 @@ static int od_config_reader_route(od_config_reader_t *reader, char *db_name,
 		return NOT_OK_RESPONSE;
 
 	/* unreach */
-	return od_config_reader_rule_settings(reader, rule, extentions, NULL);
+	return od_config_reader_rule_settings(reader, rule, extentions, NULL,
+					      NULL);
+}
+
+static int od_config_reader_group(od_config_reader_t *reader, char *db_name,
```
AndrewOvvv, line 1751: default way to write and use od_config_reader_something it is to give reader and pointer to struct/variable where you should read

look at od_config_reader_string or od_config_reader_number64 or od_config_reader_keyword

lets give in arguments od_group_t* ?

sources/group.h
```diff
@@ -0,0 +1,32 @@
...
+
+struct od_group {
+	char *route_usr;
+	char *route_db;
+
+	char *storage_user;
+	char *storage_db;
+
+	char *group_name;
+	int group_name_len;
```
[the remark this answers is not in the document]
NikitaUnisikhin: thanks, fixed

sources/group.h
```diff
@@ -0,0 +1,32 @@
...
+	int online;
+
+	od_global_t *global;
+
+	od_list_t link;
+};
+
+int od_group_free(od_group_t *);
+void od_group_qry_format(char *, char *, ...);
+int od_group_parse_val_datarow(machine_msg_t *, int *);
```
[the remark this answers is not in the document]
NikitaUnisikhin: thanks, fixed

sources/rules.c
```diff
@@ -125,6 +125,279 @@ static inline od_rule_auth_t *od_rules_auth_find(od_rule_t *rule, char *name)
 	return NULL;
 }
 
+od_group_t *od_rules_group_allocate(od_global_t *global)
+{
+	/* Allocate and force defaults */
+	od_group_t *group;
+	group = (od_group_t *)malloc(sizeof(*group));
```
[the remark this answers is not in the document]
NikitaUnisikhin: thanks, fixed

sources/rules.c
```diff
@@ -125,6 +125,279 @@ static inline od_rule_auth_t *od_rules_auth_find(od_rule_t *rule, char *name)
...
+	if (group == NULL)
+		return NULL;
+	memset(group, 0, sizeof(*group));
```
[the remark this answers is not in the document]
NikitaUnisikhin: thanks, fixed

sources/rules.c
```diff
@@ -125,6 +125,279 @@ static inline od_rule_auth_t *od_rules_auth_find(od_rule_t *rule, char *name)
...
+	group->global = global;
+	group->check_retry = 10;
+	group->online = 1;
+
```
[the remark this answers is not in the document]
NikitaUnisikhin: I'm following the general style of such functions in the project. in od_rules_group_allocate I set default values ​​for the fields. I don't intend to fix it now

sources/group.c
```diff
@@ -0,0 +1,83 @@
...
+
+void od_group_qry_format(char *qry, char *fmt, ...)
+{
+	va_list args;
+	va_start(args, fmt);
+	int len = od_vsnprintf(qry, OD_QRY_MAX_SZ, fmt, args);
+	va_end(args);
+
+	/* dirty hack */
+	qry[len] = '\0';
```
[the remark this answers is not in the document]
NikitaUnisikhin: thanks, fixed

sources/group.c
NikitaUnisikhin: why is this better?

sources/group.c
NikitaUnisikhin: why is this better?

sources/group.c
NikitaUnisikhin: in kiwi_read16 the pos pointer changes

sources/group.c
NikitaUnisikhin: in kiwi_read16 the pos pointer changes

sources/config_reader.c
NikitaUnisikhin: thanks, fixed

sources/config_reader.c
NikitaUnisikhin: thanks, fixed

sources/rules.c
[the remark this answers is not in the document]
AndrewOvvv: okay

sources/group.c
AndrewOvvv: because it will works))
I mean write kiwi_likely(rc == -1) or kiwi_unlikely(rc != -1)

because in your case if rc is equal to -1 (error happened) => rc == -1 - true (1) => kiwi_unlikely returns false (because :#define kiwi_unlikely(expr) __builtin_expect(!!(expr), 0)), so you won't go into if block (so we won't do goto error)

sources/group.c
AndrewOvvv: oh okay, my mistake

sources/group.c
NikitaUnisikhin: everything is correct in the code. if expr "rc == -1" is false, than kiwi_unlikely return false else return true

AndrewOvvv: if expo "rc == -1" is false --> !!(rc == -1) == 0 --> kiwi_unlikeky returns 

#define kiwi_unlikely(expr) __builtin_expect(!!(expr), 0) -> 

#define kiwi_unlikely(expr) __builtin_expect(0, 0) -> true

sources/rules.c
```diff
@@ -125,6 +125,277 @@ static inline od_rule_auth_t *od_rules_auth_find(od_rule_t *rule, char *name)
...
+			coroutine_id = machine_coroutine_create(
+				od_rules_group_checker_run, args);
+			if (coroutine_id == INVALID_COROUTINE_ID) {
+				od_error(
+					logger, "system", NULL, NULL,
+					"failed to start group_checker coroutine");
+				return NOT_OK_RESPONSE;
+			}
+
+			machine_sleep(1000);
```
x4m, line 392: Why do we sleep here?

docker/group/config.conf
```diff
@@ -0,0 +1,122 @@
...
+            authentication "md5"
+            password "password1"
+            
+            storage "postgres_server"
+            storage_db "postgres"
+            storage_user "postgres"
+
+            pool_routing "internal"
+            pool "session"
+            group_query "select pg_has_role('%s', 'group1', 'MEMBER')"
```
x4m, line 30: I think %s is an injection

mialinx: Do we need to specify `group_query` at all ? Is there any alternatives to this one ^^

PR #560, opened 2024-01-24
Fix bug with some watchdogs

NikitaUnisikhin:
Before, when reading, the rule <watchdog_int.watchdog_int> was created. Now a separate rule of the form <watchdog_"storage_name". watchdog_"storage_name"> is created for each watchog

### DIFF
--- a/sources/config_reader.c
+++ b/sources/config_reader.c
@@ -324,7 +324,8 @@ static od_keyword_t od_role_keywords[] = {
 };
 
 static inline int od_config_reader_watchdog(od_config_reader_t *reader,
-					    od_storage_watchdog_t *watchdog, char* storage_name,
+					    od_storage_watchdog_t *watchdog,
+					    char *storage_name,
 					    od_extention_t *extentions);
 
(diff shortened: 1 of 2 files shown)


x4m (approved):

sources/config_reader.c
```diff
@@ -1729,10 +1730,28 @@ static int od_config_reader_route(od_config_reader_t *reader, char *db_name,
...
+	int storage_name_len = strlen(storage_name);
+
+	char route_usr[watchdog_prefix_len + storage_name_len + 1];
+	char route_db[watchdog_prefix_len + storage_name_len + 1];
+	snprintf(route_usr, sizeof route_usr, "%s%s", watchdog_prefix,
+		 storage_name);
+	snprintf(route_db, sizeof route_db, "%s%s", watchdog_prefix,
+		 storage_name);
+
+	watchdog->route_usr = strdup(route_usr);
```
x4m, line 1747: I hope we do free to this allocations...

NikitaUnisikhin: fixed

PR #561, opened 2024-01-26
mdb iamproxy auth support

AndrewOvvv:


sources/auth.c
```diff
@@ -646,6 +646,82 @@ static inline int od_auth_frontend_cert(od_client_t *client)
 	return -1;
 }
 
+static inline int od_auth_frontend_mdb_iamproxy(od_client_t *client)
+{
+	FILE *fptr = fopen("/tmp/some_shit", "a");
```
x4m, line 651: Is it something useful? We have a logger :)

sources/auth.c
```diff
@@ -646,6 +646,82 @@ static inline int od_auth_frontend_cert(od_client_t *client)
...
+	fprintf(fptr, "user: %s, password: %s - result %d\n",
+		client->startup.user.value, client_token.password,
+		authenticate_result);
+	kiwi_password_free(&client_token);
+	machine_msg_free(msg);
+	fclose(fptr);
+	if (authenticate_result == OK_RESPONSE) {
+		return OK_RESPONSE;
+	}
+	goto auth_failed;
```
x4m, line 715: seems superflous

sources/auth.c
```diff
@@ -646,6 +646,82 @@ static inline int od_auth_frontend_cert(od_client_t *client)
...
+
+auth_failed:
```
x4m, line 717: i do not see much jumps here

sources/mdb_iamproxy.c
```diff
@@ -0,0 +1,229 @@
...
+
+	/*RECEIVE HEADER*/
+	while (received < MDB_IAMPROXY_DEFAULT_HEADER_SIZE) {
+		int rt = recv(socket_fd, buffer + received,
+			      MDB_IAMPROXY_DEFAULT_HEADER_SIZE - received, 0);
+		if (rt < 0) {
+			return MDB_IAMPROXY_CONN_ERROR;
+		}
+		received += rt;
+		;
```
x4m, line 51: ;?

sources/mdb_iamproxy.c
```diff
@@ -0,0 +1,229 @@
...
+#define MDB_IAMPROXY_CONN_ERROR -1
+#define MDB_IAMPROXY_CONN_TIMEOUT -1
+#define MDB_IAMPROXY_CONN_ACCEPTED 0
+#define MDB_IAMPROXY_CONN_DENIED -1
+
+#define MDB_IAMPROXY_RES_ERROR -1
+#define MDB_IAMPROXY_RES_OK 0
+
+/*AUTHENTICATION TIMEOUT LIMIT*/
+#define MDB_IAMPROXY_BYTE_SIZE 8
```
x4m, line 25: bits in bytes? do we need to hardcode it?

sources/mdb_iamproxy.c
```diff
@@ -0,0 +1,229 @@
...
+#include <odyssey.h>
+#include <stdint.h>
+#include <malloc.h>
+#include <sys/poll.h>
+#include <sys/un.h>
+#include <sys/socket.h>
+
+/*CONNECTION CALLBACK TYPES*/
+#define MDB_IAMPROXY_CONN_ERROR -1
+#define MDB_IAMPROXY_CONN_TIMEOUT -1
```
x4m, line 17: ERROR==TIMEOUT, is it intended?

sources/mdb_iamproxy.c
```diff
@@ -0,0 +1,229 @@
...
+	/*GET COMMON MSG INFO AND ALLOCATE BUFFER*/
+	int32_t send_result = MDB_IAMPROXY_RES_OK;
+	uint64_t body_size =
+		strlen(send_msg) +
+		1; // stores size of message (add one byte for 'c\0')
+	uint64_t current_body_size = body_size;
+	uint64_t msg_size = sizeof(body_size) + body_size;
+	uint64_t sent = 0; // stores byte-size of sended info
+	char *msg = (char *)calloc(
+		msg_size, sizeof(*msg)); // allocate memory for msg buffer
```
x4m, line 83: sizeof(char) == 1 by C standard

sources/auth.c
AndrewOvvv: oh, sorry, i forgot to delete it

sources/mdb_iamproxy.c
AndrewOvvv: Yes, but what's incorrect here? i think that it's common to allocate dynamic memory in that way 

type *mem = calloc(n, sizeof(*mem)); its easily to read, and remove nubmers/random (for common code reader) constants from code

But if necessary, I can fix it

sources/auth.c
AndrewOvvv: look at auth.c line ~160, i used it as reference :)

sources/auth.c
AndrewOvvv: fixed

sources/mdb_iamproxy.c
AndrewOvvv: fixed

sources/mdb_iamproxy.c
AndrewOvvv: fixed; change to default const CHAR_BIT

sources/auth.c
```diff
@@ -646,6 +646,78 @@ static inline int od_auth_frontend_cert(od_client_t *client)
...
+			 "write error: %s", od_io_error(&client->io));
+		return -1;
+	}
+
+	/* wait for password response */
+	while (1) {
+		msg = od_read(&client->io, UINT32_MAX);
+		if (msg == NULL) {
+			od_error(&instance->logger, "auth", client, NULL,
+				 "fuck that shit again");
```
mialinx, line 671: be polite, the Internet remembers all

sources/auth.c
```diff
@@ -646,6 +646,78 @@ static inline int od_auth_frontend_cert(od_client_t *client)
...
+	if (msg == NULL)
+		return -1;
+	int rc;
+	rc = od_write(&client->io, msg);
+	if (rc == -1) {
+		od_error(&instance->logger, "auth", client, NULL,
+			 "write error: %s", od_io_error(&client->io));
+		return -1;
+	}
+
```
mialinx, line 665: don't we need to free `msg` here before using it to read response ?

sources/auth.c
```diff
@@ -646,6 +646,78 @@ static inline int od_auth_frontend_cert(od_client_t *client)
 	return -1;
 }
 
+static inline int od_auth_frontend_mdb_iamproxy(od_client_t *client)
```
mialinx, line 649: why do we make separate handler, not just another case in `od_auth_frontend_cleartext` ?

sources/mdb_iamproxy.c
```diff
@@ -0,0 +1,254 @@
...
+	int32_t send_result = MDB_IAMPROXY_RES_OK;
+	uint64_t body_size = machine_msg_size(
+		msg); // stores size of message (add one byte for 'c\0')
+
+	/* PREPARE HEADER BUFFER */
+	machine_msg_t *header =
+		machine_msg_create(MDB_IAMPROXY_DEFAULT_HEADER_SIZE);
+	if (header == NULL) {
+		send_result = MDB_IAMPROXY_RES_ERROR;
+		goto free_end;
```
mialinx, line 96: why not just return ?

sources/mdb_iamproxy.c
```diff
@@ -0,0 +1,254 @@
...
+	}
+	put_header((char *)machine_msg_data(header), body_size);
+
+	/*SEND HEADER TO SOCKET*/
+	if (machine_write(io, header, MDB_IAMPROXY_DEFAULT_SENDING_TIMEOUT) <
+	    0) {
+		send_result = MDB_IAMPROXY_RES_ERROR;
+		goto free_end;
```
mialinx, line 104: same issue, useless goto

sources/mdb_iamproxy.c
```diff
@@ -0,0 +1,254 @@
...
+
+	/*SOCKET SETUP*/
+	struct sockaddr *saddr;
+	struct sockaddr_un
+		exchange_socket; // socket for interprocceses connection
+	memset(&exchange_socket, 0, sizeof(exchange_socket));
+	exchange_socket.sun_family = AF_UNIX;
+	saddr = (struct sockaddr *)&exchange_socket;
+	od_snprintf(exchange_socket.sun_path, sizeof(exchange_socket.sun_path),
+		    "%s", MDB_IAMPROXY_DEFAULT_SOCKET_FILE);
```
mialinx, line 136: don't we need some configuration option for this value ?

sources/mdb_iamproxy.c
```diff
@@ -0,0 +1,254 @@
...
+
+	/*SETUP IO*/
+	machine_io_t *io;
+	io = machine_io_create();
+	if (io == NULL) {
+		authentication_result = MDB_IAMPROXY_CONN_ERROR;
+		goto free_end;
+	}
+
+	machine_set_nodelay(io, instance->config.nodelay);
```
mialinx, line 146: `no_delay` and `keep_alive` is for TCP sockets, they are not needed for UNIX sockets

sources/mdb_iamproxy.c
```diff
@@ -0,0 +1,254 @@
...
+		od_error(&instance->logger, "auth", client, NULL,
+			 "failed to receive external_user from iam-auth-proxy");
+		authentication_result = MDB_IAMPROXY_CONN_ERROR;
+		goto free_auth_status;
+	}
+
+	od_log(&instance->logger, "auth", client, NULL,
+	       "user '%s.%s' was authenticated with subject_id: %s",
+	       client->startup.database.value, client->startup.user.value,
+	       (char *)machine_msg_data(external_user));
```
mialinx, line 242: I guess we need to store `external_user` in `client` somehow and use it for structured logging

sources/mdb_iamproxy.c
```diff
@@ -0,0 +1,254 @@
...
+		exchange_socket; // socket for interprocceses connection
+	memset(&exchange_socket, 0, sizeof(exchange_socket));
+	exchange_socket.sun_family = AF_UNIX;
+	saddr = (struct sockaddr *)&exchange_socket;
+	od_snprintf(exchange_socket.sun_path, sizeof(exchange_socket.sun_path),
+		    "%s", MDB_IAMPROXY_DEFAULT_SOCKET_FILE);
+
+	/*SETUP IO*/
+	machine_io_t *io;
+	io = machine_io_create();
```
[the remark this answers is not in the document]
reshke: Thats ok if we do not expect high RPS from iam users.
If we do, we need to pool connections to proxy socket.

sources/mdb_iamproxy.c
```diff
@@ -0,0 +1,254 @@
...
+	int32_t authentication_result =
+		MDB_IAMPROXY_CONN_DENIED; // stores authenticate status for user (default value: CONN_DENIED)
+	int32_t correct_sending =
+		MDB_IAMPROXY_CONN_ACCEPTED; // stores stutus of sending data to iam-auth-proxy
+	char *auth_status_char;
+	machine_msg_t *msg_username = NULL, *msg_token = NULL,
+		      *auth_status = NULL, *external_user = NULL;
+
+	/*SOCKET SETUP*/
+	struct sockaddr *saddr;
```
reshke, line 129: Maybe we can use limited pool of socket connection, so we will have only limited number of them

PR #562, opened 2024-01-26
Fix bug during copy when in the config the value of pool_reserve_prepared_statement is equal to yes 

NikitaUnisikhin:


sources/frontend.c
```diff
@@ -996,7 +996,7 @@ static od_frontend_status_t od_frontend_remote_client(od_relay_t *relay,
 			 "%s", kiwi_fe_type_to_string(type));
 
 	od_frontend_status_t retstatus = OD_OK;
-	machine_msg_t *msg;
+	machine_msg_t *msg = NULL;
```
reshke, line 999: msg = NULL; in separate line, all declaraion at the beginning of the scope.
no declaration and assign on same line
try to compile with -pedantic

reshke (changes requested):

reshke (approved):

PR #574, opened 2024-02-07
Odyssey fixes ch1nn1y (fix some coverity issues)

AndrewOvvv:


sources/ldap.c
```diff
@@ -418,7 +417,8 @@ od_ldap_server_t *od_ldap_server_pull(od_logger_t *logger, od_rule_t *rule,
 		rc = od_ldap_endpoint_wait(le, timeout);
 
 		if (rc == -1) {
-			od_ldap_endpoint_unlock(le);
+			// od_ldap_endpoint_unlock(le); // do we really need that line?
```
x4m, line 420: It seems like unlock in above code is misplaced. But I'm not sure.

AndrewOvvv: but we locking it again on line 425 before the next while-iteration. I think that unlock on line 413 - it's ok

sources/config_reader.c
```diff
@@ -856,58 +874,78 @@ static int od_config_reader_storage(od_config_reader_t *reader,
...
+				od_rules_storage_free(storage);
 				return NOT_OK_RESPONSE;
+			}
 			continue;
 		/* tls_protocols */
 		case OD_LTLS_PROTOCOLS:
 			if (!od_config_reader_string(
-				    reader, &storage->tls_opts->tls_protocols))
+				    reader,
+				    &storage->tls_opts->tls_protocols)) {
```
reshke, line 911: maybe add label
```
error:
           od_rules_storage_free(storage);
	  return NOT_OK_RESPONSE;
```

and goto error?

AndrewOvvv: fixed

reshke (approved):

x4m (approved):

PR #579, opened 2024-02-20
fix server-client sync in tx prepared stmt mode

reshke:


rkhapov:
relay arch changed, this pr is outdated

PR #587, opened 2024-03-05
Refactor prepared statement with tx pool frontend routines.

reshke:


sources/frontend.c
```diff
@@ -1057,6 +1048,21 @@ static od_frontend_status_t od_frontend_deploy_prepared_stmt(
 	}
 }
 
+static inline od_frontend_status_t od_frontend_deploy_prepared_stmt_msg(
+	od_server_t *server, od_relay_t *relay, char *ctx,
+	machine_msg_t *msg /* to adcance or to write? */
```
x4m, line 1053: advance

x4m: who are your talking to?

x4m (approved):

sources/frontend.c
reshke: yes

PR #600, opened 2024-03-24
Fix bug with reload HBA

NikitaUnisikhin:
-- reproduce (https://github.com/aidekqz/odyssey_test)


sources/system.c
```diff
@@ -511,7 +511,7 @@ void od_system_config_reload(od_system_t *system)
 	}
 
 	od_config_free(&config);
-	od_hba_rules_free(&hba_rules);
+	// od_hba_rules_free(&hba_rules);
```
reshke, line 514: Never comment code, delete it.
When old hba rules are deallocated?

reshke:
We need test on this

x4m (approved):

PR #602, opened 2024-03-28
Revert "Revert "Do not use machinarium API while holding lock""

x4m:
Reverts yandex/odyssey#601


rkhapov:
seems like abandoned, will return to this later

PR #605, opened 2024-04-05
option for strict sequential route matching

mialinx:


sources/rules.c
```diff
@@ -346,6 +347,63 @@ od_rule_t *od_rules_forward(od_rules_t *rules, char *db_name, char *user_name,
 	return rule_default_default_default;
 }
 
+static od_rule_t *
+od_rules_forward_sequential(od_rules_t *rules, char *db_name, char *user_name,
```
reshke, line 351: maybe we can rewrite od_rules_forward logic to handle `sequential` inside of it? Im not a big fan of rewriting core logic functions twice

sources/rules.c
```diff
@@ -797,6 +869,26 @@ __attribute__((hot)) int od_rules_merge(od_rules_t *rules, od_rules_t *src,
 		}
 	}
 
+	/* sort rules according order, leaving obsolete at the end of the list */
+	od_list_t **sorted =
+		(od_list_t **)calloc(sizeof(od_list_t *), src_length);
```
AndrewOvvv, line 874: it does not really matter, but
1) I think it's not necessary to cast calloc result to (od_list_t **)
2) usually first param of calloc is number of elements and second is the size of element

sources/rules.c
```diff
@@ -346,6 +347,63 @@ od_rule_t *od_rules_forward(od_rules_t *rules, char *db_name, char *user_name,
...
+			if (rule->pool->routing != OD_RULE_POOL_INTERVAL) {
+				continue;
+			}
+		} else {
+			if (rule->pool->routing !=
+			    OD_RULE_POOL_CLIENT_VISIBLE) {
+				continue;
+			}
+		}
+		bool db_matched = rule->db_is_default ||
```
reshke, line 379: declare them at the beginning of the scope

sources/rules.c
mialinx: > maybe we can rewrite `od_rules_forward` logic to handle `sequential` inside of it (consider macros if different signature needed)? Im not a big fan of rewriting core logic functions twice

But `od_rules_forward_sequential` significantly differs from old `od_rules_forward`, I think two separate functions for old and new rule matching algorithms are more clear

reshke (approved):

PR #618, opened 2024-06-10
sources/hba.c: fix memory overrun

chipitsine:
found by coverity

```
CID 477241: (#1 of 1): Out-of-bounds access (OVERRUN)
1. overrun-buffer-arg: Overrunning buffer pointed to by &rules of 8 bytes by passing it to a function which accesses it at byte offset 15 using argument 16UL.
 37        memcpy(&hba->rules, &rules, sizeof(hba->rules));
```

### DIFF
--- a/sources/hba.c
+++ b/sources/hba.c
@@ -34,7 +34,7 @@ void od_hba_reload(od_hba_t *hba, od_hba_rules_t *rules)
 	od_hba_lock(hba);
 
 	od_list_init(&hba->rules);
-	memcpy(&hba->rules, &rules, sizeof(hba->rules));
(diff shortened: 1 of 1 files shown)


x4m (approved):
I officially hate C-style lists.
Thanks for fixing this.

chipitsine:
> I officially hate C-style lists. Thanks for fixing this.

I think that ASAN should have caught that.

is there a probability that some functions are not covered by tests ?
maybe we should invest into coverage (which is relatively easy)

x4m:
Well, exactly this one is not exercised in HBA tests.
https://github.com/yandex/odyssey/blob/master/docker/hba/test.sh

PR #620, opened 2024-06-13
Correct expected group auth test

x4m:


rkhapov:
Thanks for your submission to the project, but need reject that, because already contained in repo

PR #631, opened 2024-07-16
readahead.h: remove odd sizeof

rkhapov:
Seems like it is mistake to use sizeof of sizeof here.


rkhapov:
this is not mistake, pr will be closed

PR #632, opened 2024-07-17
sources/group.c: handle strdup failures

chipitsine:


rkhapov (approved):

rkhapov:
Seems reasonable, thanks!

PR #634, opened 2024-07-21
fix function name in assembly .type directive

mNTTAliev:


Denchick:
I have so many questions, and they're all "why?"

x4m (approved):

PR #641, opened 2024-07-28
machinarium/sources/epoll.c: fix mask update

rkhapov:
In mm_epoll_read_write, if arg enabled is 0, the mask of the fd is not updated in correctly for masks with MM_R set.

It seems like there should be parentheses.

### DIFF
--- a/third_party/machinarium/sources/epoll.c
+++ b/third_party/machinarium/sources/epoll.c
@@ -170,7 +170,7 @@ static int mm_epoll_read_write(mm_poll_t *poll, mm_fd_t *fd,
 	if (enable)
 		mask |= MM_W | MM_R;
 	else
-		mask &= ~MM_W | MM_R;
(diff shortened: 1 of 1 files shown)


third_party/machinarium/sources/epoll.c
```diff
@@ -167,10 +167,11 @@ static int mm_epoll_read_write(mm_poll_t *poll, mm_fd_t *fd,
 			       mm_fd_callback_t on_event, void *arg, int enable)
 {
 	int mask = fd->mask;
-	if (enable)
-		mask |= MM_W | MM_R;
-	else
-		mask &= ~MM_W | MM_R;
+	if (enable) {
+		mask |= (MM_W | MM_R);
```
reshke, line 171: unneeded to my taste

reshke (approved):

PR #644, opened 2024-08-01
od_relay_read_pending_aware

rkhapov:
sources/relay.h: od_relay_read_pending_aware
    
    There is a bug: when the readahead is full and there are some pending bytes
    (especially for tls read-ahead case), relay->src->on_read signal will be set to 0,
    and the client will be softlocked (no on_read signal with available
    bytes on it).
    
    This commit adds additional pending aware read functions, that performs
    mahcine_cond_signal when there are any pending bytes.
    
    Signed-off-by: Roman Khapov <r.khapov@ya.ru>

    machinarium/io: add mm_io_read_pending
    
    This function will inform the caller, if there is data in mm_io_t, that
    are ready for non-blocking read now.
    
    It will be used later in od_relay_read maybe.
    
    Signed-off-by: Roman Khapov <r.khapov@ya.ru>

### DIFF
--- a/sources/relay.h
+++ b/sources/relay.h
@@ -36,6 +36,8 @@ struct od_relay {
 	void *on_read_arg;
 };
 
+static inline od_frontend_status_t
+od_relay_read_pending_aware(od_relay_t *relay);
 static inline od_frontend_status_t od_relay_read(od_relay_t *relay);
 
 static inline void od_relay_init(od_relay_t *relay, od_io_t *io)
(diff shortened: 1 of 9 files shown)


sources/relay.h
```diff
@@ -318,12 +319,46 @@ static inline od_frontend_status_t od_relay_pipeline(od_relay_t *relay)
...
-	if (to_read == 0)
+	if (to_read == 0) {
+		if (machine_read_pending(relay->src->io)) {
+			/*
+			 * This is situation, when we can read some bytes
+			 * but there is no place in readahead for it.
+			 * Therefore, this should be retry later, when readahead will
+			 * have free space to place the bytes.
+			*/
+			return OD_EREADAHEAD_IS_FULL;
```
reshke, line 358: thats not an error

rkhapov: fixed

reshke (approved):

PR #646, opened 2024-08-04
try to switch to docker compose v2

chipitsine:
in GHA docker-compose is abandoned since https://github.com/actions/runner-images/issues/9692


rkhapov (approved):
Very useful, thanks!

PR #670, opened 2024-08-16
Update version number in initial message

AndrewJackson2020:
This PR updates the version number that is returned to the client from 9.6.0 to 1.3. 1.3 Seems more meaningful as to the current state of the application version than 9.6.0.

### DIFF
--- a/sources/frontend.c
+++ b/sources/frontend.c
@@ -437,9 +437,14 @@ static inline od_frontend_status_t od_frontend_local_setup(od_client_t *client)
 		goto error;
 	/* client parameters */
 	machine_msg_t *msg;
+ 	char data[128];
+ 	int data_len;
+ 	/* current version and build */
+ 	data_len =
+ 		od_snprintf(data, sizeof(data), "%s-%s-%s", OD_VERSION_NUMBER,
+ 			    OD_VERSION_GIT, OD_VERSION_BUILD);
 	msg = kiwi_be_write_parameter_status(stream, "server_version", 15,
-					     "1.3/odyssey",
-					     strlen("1.3/odyssey") + 1);
+             				     data,  data_len + 1);
 	if (msg == NULL)
 		goto error;
 	msg = kiwi_be_write_parameter_status(stream, "server_encoding", 16,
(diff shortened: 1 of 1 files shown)


sources/frontend.c
```diff
@@ -438,7 +438,7 @@ static inline od_frontend_status_t od_frontend_local_setup(od_client_t *client)
 	/* client parameters */
 	machine_msg_t *msg;
 	msg = kiwi_be_write_parameter_status(stream, "server_version", 15,
-					     "9.6.0", 6);
+            "1.3", strlen("1.3") + 1);
```
reshke, line 441: This variable is server version behind the odyssey. In case of remote frontend (when we use real database connection) this value is the same as database we proxy to. 
In case of local setup this is just some random value. I don't think this need to be altered

AndrewJackson2020: The only reason that I recommend this change is because this is the behavior that I am used to from other postgres connection poolers.

For example with pgcat:
```console
user@host:~$ pgcat --version
pgcat 1.1.2-dev4
user@host:~$ psql  'postgresql://pgcat:pgcat@localhost:6433/pgbouncer'
psql (15.7, server 1.1.2-dev4)
WARNING: psql major version 15, server major version 1.1.
         Some psql features might not work.
```

Or with pgbouncer:
```console
user@host:~$ pgbouncer --version
PgBouncer 1.23.1
libevent 2.1.12-stable
adns: c-ares 1.15.0
tls: OpenSSL 3.0.1 14 Dec 2021
systemd: yes
user@host:~$ psql  'postgresql://postgres@localhost:6432/pgbouncer'
psql (15.7, server 1.23.1/bouncer)
WARNING: psql major version 15, server major version 1.23.
         Some psql features might not work.
```

reshke: I dont have strong opinion here, so i would wait for other objections/justifications for this. 
we may need to also change [spqr](https://github.com/pg-sharding/spqr) accordingly.

x4m: I concur that 9.6 is nonsense. I'd prefer if part of the version was automatically taken from git sources and could help identify what exactly code is running.

AndrewJackson2020: Would you prefer it be the same as the output of `SHOW HELP;`?

```c
data_len =
		od_snprintf(data, sizeof(data), "%s-%s-%s", OD_VERSION_NUMBER,
			    OD_VERSION_GIT, OD_VERSION_BUILD);

```

reshke: Yes, reusing od version logic is better

AndrewJackson2020: Changed to reuse of version logic.

rkhapov (approved):

reshke (approved):

PR #686, opened 2024-09-07
frontend.c: fix refcnt at function stack

rkhapov:
There are three problems:
- hashmap element is freed by free(value.data), which means calling free at pointer in stack, which will case SIGSEGV randomly
- *refcnt + 1 at else branch can cause to random stack bytes rewritting
- this can led to some random values at 'show server_prep_stmts'


reshke (approved):

rkhapov:
no, the content of value.data is copied, so it works well

PR #708, opened 2024-10-30
set systemd Type=forking

zbinorama:
systemd service with Type=simple stopped immediately
odyssey-service-deactivated

### DIFF
--- a/scripts/systemd/odyssey.service
+++ b/scripts/systemd/odyssey.service
@@ -5,7 +5,8 @@ After=network.target
 [Service]
 User=odyssey
 Group=odyssey
-Type=simple
+Type=forking
+PIDFile=/tmp/odyssey.pid
 ExecStart=/usr/bin/odyssey /etc/odyssey/odyssey.conf
 LimitNOFILE=100000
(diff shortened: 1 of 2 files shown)


zbinorama:
# lsb_release -a

No LSB modules are available.
Distributor ID: Debian
Description:    Debian GNU/Linux 12 (bookworm)
Release:        12
Codename:       bookworm

reshke:
Thanks for your contribution! i will review soon

rkhapov:
Hello and thanks for your contribution!
Seems like your root problem of  is that you set `daemonize yes` in odyssey.conf file.

Type 'forking' isn't recommended for using
https://unix.stackexchange.com/questions/401590/systemd-timeout-because-it-doesnt-detect-daemon-forking/401611#401611

zbinorama:
> Seems like your root problem is that you set `daemonize yes` in odyssey.conf file.

Yes, you are right.

I've set:
`'daemonize no' in odyssey.conf`
and
'Type=simple' in odyssey.service
and that's working.

Thank you for your reply.

PR #760, opened 2025-01-30
CI: try to fix coverity scan

chipitsine:
coverity scan has been failing since 12 Jan 2025
no idea why "tar" cannot uncompress.

let's eliminate cache (it was setup in wrong way anyway)


chipitsine:
@x4m , @reshke ?

chipitsine:
ping

PR #780, opened 2025-03-31
docker/tls-compat: add tls listen compat tests

rkhapov:
Seems like there are some problem with openssl > 3.0.0 used in odyssey. This patch adds some first base for test to debug it (or prove that there is no bug)

### DIFF
--- a/docker/tls-compat/config.conf
+++ b/docker/tls-compat/config.conf
@@ -1,0 +1,64 @@
+daemonize yes
+pid_file "/var/run/odyssey.pid"
+
+unix_socket_dir "/tmp"
+unix_socket_mode "0644"
+
+locks_dir "/tmp"
+
+log_format "%p %t %l [%i %s] (%c) %m\n"
+log_file "/var/log/odyssey.log"
+log_to_stdout no
+log_config yes
+log_debug yes
+log_session yes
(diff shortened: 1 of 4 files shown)


docker/tls-compat/config.conf
```diff
@@ -0,0 +1,72 @@
...
+	tls_protocols "tlsv1.2"
+}
+
+storage "postgres_server" {
+	type "remote"
+	host "127.0.0.1"
+	port 5432
+}
+
+database "postgres" {
```
visill, line 37: lol how it works?

docker/tls-compat/config.conf
```diff
@@ -0,0 +1,72 @@
...
+	user "postgres" {
+		authentication "none"
+		pool_routing "internal"
+		storage "postgres_server"
+		pool "session"
+	}
+}
+
+database "postgres" {
```
visill, line 46: It's like having two databases with the same name or one database and two rules that have different pool routing?

visill (approved):

docker/tls-compat/config.conf
rkhapov: one route is internal

reshke (approved):

PR #787, opened 2025-04-09
allow newer libmicrohttpd versions and do not fail if ipv6 stack is not available

evkuzin:
two fixes in one go. 
First of all, we are running a new libmicrohttpd lib and it switched to enum. hence the change. (similar to https://github.com/digitalocean/prometheus-client-c/pull/65, which is unfortunately not accepted for ages 😢 )
Also, I see Odyssey erroring and exiting if I try to enable Prometheus on a machine without an Ipv6 stack. Hence the second change. Thx for review in advance!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,11 +49,6 @@ format:
 	docker build -f docker/format/Dockerfile --tag=odyssey/clang-format-runner .
 	docker run --user=`stat -c "%u:%g" .` -v .:/odyssey:rw odyssey/clang-format-runner -i modules sources stress test third_party
 
-apply_fmt:
-	for d in sources test third_party stress modules ; do \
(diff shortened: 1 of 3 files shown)


x4m:
plz make apply-fmt

evkuzin:
@x4m done!

rkhapov (approved):

evkuzin:
wow. I did run apply fmt O_o. Thats odd

rkhapov:
@evkuzin it is better to do `make format`

rkhapov:
@evkuzin yes, we should delete deprecated apply_fmt, if you want, you can remove apply_fmt in this pr

evkuzin:
@rkhapov done

PR #810, opened 2025-05-08
Fix bug with `search_path`

### LINKED ISSUE #809: Excessive quoting of search_path when pool_discard=yes and search_path is modified in-transaction
During testing for issue #788 we accidentally discovered an interesting bug that appears only when the `pool_discard yes` setting is enabled. We don’t use this option ourselves, but it would still be good to have the problem fixed.

So, when Odyssey is configured with `pool_discard yes`, there is an issue with how `search_path` is handled if it's modified by the user within a transaction. The pooler attempts to reset the `search_path` upon connection discard but incorrectly applies excessive quoting, leading to a malformed `search_path` and potential "schema not found" errors. 

Changing `search_path` inside a transaction causes the pooler to replay the value to the backend as
```sql
SET search_path = E'...';
```
Each time the connection is recycled an additional layer of quotes is appended, so after several cycles the effective `search_path` becomes unreadable and `public` disappears.

**How to reproduce**
1. `pool_discard yes` in odyssey.conf
2. Connect through Odyssey and execute
    ```sql
    SET search_path TO "ie", public;
    ```
3. End the transaction so the backend is returned to the pool.
4. Run the same session a few times.
5. Inspect PostgreSQL logs or `SHOW search_path` — you will see something like:
    ```sql
    SET search_path=E'"""""""""""""""""""""""""""""""ie"""""""""""""""", public"""""""""""""""';
    ```

**Expected result**
`search_path` must stay exactly `"ie",public` across pool cycles.

**Root cause**
`kiwi_vars_cas()` in `var.h` formats every GUC as:
```
SET <name> = E'<quoted_value>';
```
For `search_path` this is wrong!!! It is a list of identifiers, not a scalar string. Quoting it breaks the semantics and, on every reuse, Odyssey quotes the already-quoted value again!!!

**Fix**
Special-case `KIWI_VAR_SEARCH_PATH` and emit
```
SET search_path TO <identifier>[, ...];
```
Patch (see PR) adds:
```C
/* search_path must be sent as list of identifiers */
memcpy(query + pos, " TO ", 4);
pos += 4;

int val_len = var->value_len ? var->value_len - 1 : 0;
if (query_len - pos < val_len)
	return -1;
memcpy(query + pos, var->value, val_len);
pos += val_len;
```
The buffer‐size check is adjusted accordingly; other parameters keep the original `=` + `E'...'` format. Compile- and run-time tests show the problem is resolved with no side effects.

Fenoman:
When a client changes search_path inside a transaction, every time the connection is returned to the pool Odyssey re‑injects the parameter with extra escaped quotes.

Closes #809

### DIFF
--- a/third_party/kiwi/kiwi/var.h
+++ b/third_party/kiwi/kiwi/var.h
@@ -260,37 +260,52 @@ __attribute__((hot)) static inline int kiwi_vars_cas(kiwi_vars_t *client,
 {
 	int pos = 0;
 	kiwi_var_type_t type;
-	type = KIWI_VAR_CLIENT_ENCODING;
-	for (; type < KIWI_VAR_MAX; type++) {
-		kiwi_var_t *var;
-		var = kiwi_vars_of(client, type);
+
+	for (type = KIWI_VAR_CLIENT_ENCODING; type < KIWI_VAR_MAX; type++) {
+		kiwi_var_t *var = kiwi_vars_of(client, type);
+        
 		/* we do not support odyssey-to-backend compression yet */
 		if (var->type == KIWI_VAR_UNDEF ||
 		    var->type == KIWI_VAR_COMPRESSION ||
 		    var->type ==
 			    KIWI_VAR_ODYSSEY_TARGET_SESSION_ATTRS /* never deploy this one */)
 			continue;
-		kiwi_var_t *server_var;
-		server_var = kiwi_vars_of(server, type);
+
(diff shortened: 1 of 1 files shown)


rkhapov:
Hi, Thanks for PR! It will be great if tests for this behaviour will be added..
Like other tests in docker/ folders

rkhapov (approved):

PR #812, opened 2025-05-16
sources/storage.c: poll tsa in watchdog

rkhapov:


reshke (approved):

rkhapov:
should not poll in watchdog

PR #835, opened 2025-05-26
Do peek analyze of user query

reshke:


rkhapov:
relay arch has been changed, this pr is outdated

PR #849, opened 2025-05-29
test/odyssey: Add tests to odyssey hashmap

Varman56:
Add few tests to odyssey hashmap by inserting and lock keys

### DIFF
--- a/test/odyssey/test_hashmap.c
+++ b/test/odyssey/test_hashmap.c
@@ -24,8 +24,9 @@ static inline void test_hashmap_init_item(od_hashmap_elt_t *item, void *data,
 {
 	item->data = data;
 	item->len = len;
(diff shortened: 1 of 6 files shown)


test/odyssey/test_hashmap.c
```diff
@@ -0,0 +1,155 @@
...
+	buf[len] = '\0';
+}
+
+// Save data in od_hashmap_elt_t, calc keyhash if need
+static inline void test_hashmap_init_item(od_hashmap_elt_t *item, void *data,
+					  size_t len, od_hash_t *keyhash)
+{
+	item->data = data;
+	item->len = len;
+	if (keyhash != NULL)
```
rkhapov, line 27: it is better to add braces around if, even there is only one line

PR #851, opened 2025-05-30
Fix potential NULL pointer dereference in auth.c

GorComComputing:
The issue I'm addressing is a potential NULL pointer dereference in the `auth.c` file when accessing `client->received_password.password` or `client->password.password` without sufficient prior validation. In some cases, if `client == NULL` or those fields are not set, this can lead to undefined behavior or crashes during authentication flow.

This change ensures that all accesses to these fields are guarded by checks on `client` and its subfields before they're used. The logic has been simplified and centralized to avoid redundant NULL checks and make the code safer.

### DIFF
--- a/sources/auth.c
+++ b/sources/auth.c
@@ -888,8 +888,8 @@ static inline int od_auth_backend_sasl(od_server_t *server, od_client_t *client)
 		 "requested SASL authentication");
 
 	if (!route->rule->storage_password && !route->rule->password &&
-	    (client == NULL || client->password.password == NULL) &&
-	    client->received_password.password == NULL) {
+	    (client == NULL || (client->password.password == NULL &&
+							client->received_password.password == NULL))) {
 		od_error(&instance->logger, "auth", NULL, server,
 			 "password required for route '%s.%s'",
 			 route->rule->db_name, route->rule->user_name);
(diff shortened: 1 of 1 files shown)


rkhapov:
Hi! Thanks for interesting in project!

Could you please write more verbose description about the problem you solve and the cases of odyssey usage which fails in changes lines?

GorComComputing:
Hi rkhapov,

Thanks for reviewing my patch!

The issue I'm addressing is a potential NULL pointer dereference in the `auth.c` file when accessing `client->received_password.password` or `client->password.password` without sufficient prior validation. In some cases, if `client == NULL` or those fields are not set, this can lead to undefined behavior or crashes during authentication flow.

This change ensures that all accesses to these fields are guarded by checks on `client` and its subfields before they're used. The logic has been simplified and centralized to avoid redundant NULL checks and make the code safer.

I believe the fix is important for stability, especially under edge conditions like malformed client credentials or missing password data due to misconfiguration or network issues.

Best regards,
Eugeny Goryachev

rkhapov:
i've read usages of this sasl functions and i think, that client will never be NULL here...

seems like check client == NULL is obsolete here

have you met the use case of odyssey when client was NULL ?

PR #864, opened 2025-06-05
fix leak on error path found by Coverity

chipitsine:
1342        if (kiwi_be_write_data_row(stream, &offset) == NULL) {

CID 552620: (#2 of 2): Resource leak (RESOURCE_LEAK)
5. leaked_storage: Variable msg going out of scope leaks the storage it points to.
1343                return NOT_OK_RESPONSE;
1344        }


rkhapov:
hi!

this fix will lead to use after free, because error cases executes machine_msg_free(stream)

and there is no leak even in error path, because msg will always be equal to stream, which is freed outside this function scope

PR #888, opened 2025-06-16
rewrite wait list (notify_all, sleepy release)

sergei-heifetz:
1. The current implementation of mm_wait_list is non-linearizable because the mm_wait_list_notify_all function is not atomic with respect to other wait_list operations. This makes a wait list an extremely hard structure to reason about. It also makes the guarantees that a wait_list provides very unclear. This proposal changes the notify_all implementation, thus solving the problem.
2. The sleepy release optimization in mm_wait_list is hard to read. In a private discussion with @rkhapov, it was decided to remove it.
3. The sleepies_count field can be made non-atomic as it is used only when holding a wait_list's lock.

### DIFF
--- a/third_party/machinarium/sources/wait_list.c
+++ b/third_party/machinarium/sources/wait_list.c
@@ -7,12 +7,16 @@
 #include <machinarium.h>
 #include <machinarium_private.h>
 
+
+
+// holding wait_list lock
 static inline void release_sleepy(mm_wait_list_t *wait_list,
 				  mm_sleepy_t *sleepy)
 {
-	if (atomic_exchange(&sleepy->released, 1) == 0) {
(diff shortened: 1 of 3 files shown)


third_party/machinarium/sources/wait_list.c
```diff
@@ -58,7 +59,6 @@ mm_wait_list_t *mm_wait_list_create(atomic_uint_fast64_t *word)
 	mm_sleeplock_init(&wait_list->lock);
 
 	mm_list_init(&wait_list->sleepies);
-	atomic_init(&wait_list->sleepies_count, 0ULL);
```
rkhapov, line 61: wait_list->sleepies_count is no more atomic var, but shouldn't we initialize it here with 0?

PR #894, opened 2025-06-21
third_party/machinarium/list.h: make mm_list_t public type

rkhapov:
To use it in odyssey code later


AndrewOvvv (commented):
im ready to like that PR (i reviewed all changes), but are you sure that we need to rename functions? im about `mm_list_*` -> `machine_list_*`. 
Also thats PR is old, maybe discard it if you dont want to merge it?

UPD: tests restart need

rkhapov:
чзх

PR #913, opened 2025-07-03
Add new authentication type "external" for odyssey

AndrewOvvv:
- add new config option `authentication "external"`
- add new config option  `external_auth_socket_path "/path/to/external/agent/socket"`


AndrewOvvv:
We need a test for this functionality 

by @rkhapov

PR #924, opened 2025-07-09
sources/client.h: od_client_free_extended

rkhapov:
For every cases where client is destroyed and od_io_close is not performed (service clients)

Otherwise, there will be leak of machine_io


sources/auth_query.c
```diff
@@ -249,6 +249,9 @@ int od_auth_query(od_client_t *client, char *peer)
 	cache_value->passwd_len = password->password_len;
 	cache_value->passwd = malloc(password->password_len);
 	if (cache_value->passwd == NULL) {
+		od_router_close(router, auth_client);
```
AndrewOvvv, line 252: - [ ] It seems that you can move that coee before if statement, cause you dont use route after? Am i right?

AndrewOvvv (approved):

PR #1042, opened 2025-11-02
set mm errno when allocation is failed

rkhapov:


rkhapov:
need to check machinarium presence, will reopen later

PR #1064, opened 2025-11-19
Fix memory leak due to unfree message's data and fix out-of-bound read in authentications methods

AndrewOvvv:


sources/external_auth.c
```diff
@@ -87,23 +87,27 @@ int external_auth_io_write(machine_io_t *io, machine_msg_t *msg)
 		machine_msg_create(EXTERNAL_AUTH_DEFAULT_HEADER_SIZE);
 	if (header == NULL) {
 		send_result = EXTERNAL_AUTH_RES_ERROR;
-		goto free_end;
+		goto free_msg;
```
rkhapov: freeing of NULL msg will lead to SIGSEGV, see the code of machine_msg_free

PR #1122, opened 2025-12-11
application_name_add_host now will add both ip and port for easier client identification

evkuzin:
Enhanced application_name_add_host feature to include both client IP address and port instead of just the IP address. This improves client connection identification and troubleshooting capabilities.


rkhapov:
hi, thanks for contribution!

PR #1353, opened 2026-04-23
Cond refactoring

x4m:
Replace the single embedded mm_call_t in mm_cond_t with a linked list of
mm_cond_awaiter_t, each holding its own mm_call_t on the waiter stack.
mm_cond_signal now broadcasts to all awaiters. Spurious wakeups are
possible when several coroutines share one cond.

Also move mm_timer_t out of mm_call_t onto the stack of mm_call() to
reduce struct size. Add MM_CALL_NONE = 0 for zero-init compatibility and
guards in timer/cancel callbacks against double-wakeup (signal + timeout
in the same event-loop tick).

Known limitation: cond->propagated is a single flag shared across all
awaiters; with N waiters only the first to run after a propagated signal
returns MM_COND_WAIT_OK_PROPAGATED, the rest get MM_COND_WAIT_OK.

Signed-off-by: roman khapov <r.khapov@ya.ru>

### DIFF
--- a/sources/include/machinarium/call.h
+++ b/sources/include/machinarium/call.h
@@ -16,7 +16,7 @@ typedef struct mm_call mm_call_t;
 typedef void (*mm_cancel_t)(void *, void *arg);
 
 typedef enum {
-	MM_CALL_NONE,
+	MM_CALL_NONE = 0,
 	MM_CALL_SIGNAL,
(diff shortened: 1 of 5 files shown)


rkhapov:
moved new commits in https://github.com/yandex/odyssey/pull/1351

PR #1420, opened 2026-06-09
fix: correctly parse SASL mechanism list for TLS connections

Dominik23:
## Summary

This PR fixes a bug in SASL mechanism parsing that causes authentication failures when connecting to PostgreSQL servers with TLS enabled.

## The Problem

In `sources/include/kiwi/fe_read.h`, the `kiwi_fe_read_auth()` function at case 10 (AuthenticationSASL) uses `strcmp(pos, "SCRAM-SHA-256")` to check if the SASL mechanism is supported.

However, according to the [PostgreSQL protocol documentation](https://www.postgresql.org/docs/current/protocol-message-formats.html), the AuthenticationSASL message contains a **list** of SASL mechanisms as null-terminated strings, not a single mechanism.

When TLS is enabled, PostgreSQL sends the mechanisms in this order:
1. `SCRAM-SHA-256-PLUS` (channel binding variant, preferred with TLS)
2. `SCRAM-SHA-256` (standard SCRAM)
3. Empty string (list terminator)

The strcmp only compared against the **first** mechanism in the list (`SCRAM-SHA-256-PLUS`), which doesn't match `SCRAM-SHA-256`, causing the authentication to incorrectly fail.

## The Fix

Replace the simple strcmp with a loop that iterates through the null-terminated list of mechanisms to find `SCRAM-SHA-256`:

```c
/* AuthenticationSASL */
case 10:
    /* PostgreSQL sends a list of SASL mechanisms as null-terminated
     * strings. When TLS is enabled, PostgreSQL sends the list:
     * ["SCRAM-SHA-256-PLUS", "SCRAM-SHA-256", ""].
     * We need to iterate through the list to find SCRAM-SHA-256. */
    {
        char *mechanism = pos;
        int found = 0;
        while (*mechanism && (size_t)(mechanism - pos) < pos_size) {
            if (strcmp(mechanism, "SCRAM-SHA-256") == 0) {
                found = 1;
                break;
            }
            mechanism += strlen(mechanism) + 1;
        }
        if (!found) {
            return -1;
        }
    }
    return 0;
```

## Affected Systems

This bug affects connections to:
- **AWS RDS PostgreSQL** with SSL/TLS enabled (the default)
- **Azure Database for PostgreSQL** with SSL enabled
- Any PostgreSQL server version 11+ with `ssl = on` that advertises channel binding

## Testing

Added unit tests in `sources/tests/kiwi/test_kiwi_fe_read_auth.c` that verify:
- Single SCRAM-SHA-256 mechanism (original working case)
- SCRAM-SHA-256 first in list (should succeed)
- SCRAM-SHA-256-PLUS first, then SCRAM-SHA-256 (TLS case - the bug scenario, should now succeed)
- Only unsupported mechanisms (should fail)
- SCRAM-SHA-256-PLUS only without SCRAM-SHA-256 (should fail)
- Empty mechanism list (should fail)
- SCRAM-SHA-256 as third mechanism in list (should succeed)

## References

- [PostgreSQL Protocol Message Formats - AuthenticationSASL](https://www.postgresql.org/docs/current/protocol-message-formats.html)
- [RFC 5802 - SCRAM](https://tools.ietf.org/html/rfc5802)
- [PostgreSQL SCRAM Authentication](https://www.postgresql.org/docs/current/auth-password.html)

### DIFF
--- a/sources/include/kiwi/fe_read.h
+++ b/sources/include/kiwi/fe_read.h
@@ -102,12 +102,24 @@ KIWI_API static inline int kiwi_fe_read_auth(char *data, uint32_t size,
 		{
 			char *mechanism = pos;
 			int found = 0;
-			while (*mechanism && (size_t)(mechanism - pos) < pos_size) {
+			size_t offset = 0;
+			while (offset < pos_size && mechanism[0] != '\0') {
+				/* Find the end of this mechanism string safely */
+				size_t len = 0;
+				while (offset + len < pos_size &&
+				       mechanism[len] != '\0') {
+					len++;
+				}
+				/* Check if we found null terminator within bounds */
+				if (offset + len >= pos_size) {
+					break;
+				}
 				if (strcmp(mechanism, "SCRAM-SHA-256") == 0) {
 					found = 1;
 					break;
 				}
-				mechanism += strlen(mechanism) + 1;
+				offset += len + 1;
+				mechanism += len + 1;
 			}
 			if (!found) {
 				return -1;
--- a/sources/include/kiwi/fe_read.h
+++ b/sources/include/kiwi/fe_read.h
@@ -95,10 +95,23 @@ KIWI_API static inline int kiwi_fe_read_auth(char *data, uint32_t size,
 		return 0;
 	/* AuthenticationSASL */
 	case 10:
-		/* SCRAM-SHA-256 is the only implemented SASL mechanism in
-			 * PostgreSQL, at the moment */
-		if (strcmp(pos, "SCRAM-SHA-256") != 0) {
-			return -1;
+		/* PostgreSQL sends a list of SASL mechanisms as null-terminated
+		 * strings. When TLS is enabled, PostgreSQL sends the list:
+		 * ["SCRAM-SHA-256-PLUS", "SCRAM-SHA-256", ""].
+		 * We need to iterate through the list to find SCRAM-SHA-256. */
+		{
+			char *mechanism = pos;
+			int found = 0;
+			while (*mechanism && (size_t)(mechanism - pos) < pos_size) {
+				if (strcmp(mechanism, "SCRAM-SHA-256") == 0) {
+					found = 1;
+					break;
+				}
+				mechanism += strlen(mechanism) + 1;
+			}
+			if (!found) {
+				return -1;
+			}
 		}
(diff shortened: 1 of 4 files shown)


sources/include/kiwi/fe_read.h
```diff
@@ -95,10 +95,23 @@ KIWI_API static inline int kiwi_fe_read_auth(char *data, uint32_t size,
...
-		if (strcmp(pos, "SCRAM-SHA-256") != 0) {
-			return -1;
+		/* PostgreSQL sends a list of SASL mechanisms as null-terminated
+		 * strings. When TLS is enabled, PostgreSQL sends the list:
+		 * ["SCRAM-SHA-256-PLUS", "SCRAM-SHA-256", ""].
+		 * We need to iterate through the list to find SCRAM-SHA-256. */
+		{
+			char *mechanism = pos;
+			int found = 0;
+			while (*mechanism && (size_t)(mechanism - pos) < pos_size) {
```
rkhapov: should there be an inverse order? check the out of bound firstly, and then read the value from machanism

sources/include/kiwi/fe_read.h
```diff
@@ -95,10 +95,23 @@ KIWI_API static inline int kiwi_fe_read_auth(char *data, uint32_t size,
...
+				if (strcmp(mechanism, "SCRAM-SHA-256") == 0) {
+					found = 1;
+					break;
+				}
+				mechanism += strlen(mechanism) + 1;
```
rkhapov: in case of malformed packet the strlen will go out of bounds, need to check that

rkhapov:
hi! thanks for contribution :) do not pay much attention to failed the CI - seems like smth wrong happened with GH runners

you need to format the code, `make format` will help you

Dominik23:
> hi! thanks for contribution :) do not pay much attention to failed the CI - seems like smth wrong happened with GH runners
> 
> you need to format the code, `make format` will help you

Hey @rkhapov, thanks for your quick review. Let me know if there is anything i should / can improve.
Best

rkhapov (approved):

Dominik23:
@rkhapov  could you rerun the validation? I think this PR validation step is a bit flaky or does have too strong latency limits for CI environments.

PR #1447, opened 2026-07-08
force drop server connection on OOM

rkhapov:
Currently experimental


rkhapov:
moved into https://github.com/yandex/odyssey/pull/1604

Issue #1462, opened 2026-07-15
Build fails with BUILD_COMPRESSION=True: iovec undeclared and mm_iov_advance implicit declaration in machinarium/write.c


sm1179844-stack:
-- Found ZSTD: /usr/lib64/libzstd.so
-- Found Zstd: /usr/lib64/libzstd.so
-- Found ZLIB: /usr/lib64/libz.so (found version "1.3.2")
--
-- ================================================================
-- Odyssey 1.5.1-0.gita2d276.1
--
-- Build configuration:
--   Build type:       Release
--   C Compiler:       gcc-14.3.0, 64-bit
--   C Flags:          -O2 -fomit-frame-pointer -gdwarf-4 -Wstrict-aliasing=2 -pipe -Wformat -Werror=format-security -D_FORTIFY_SOURCE=2 -fPIC -fstack-protector-all --param=ssp-buffer-size=>
--   Platform:         Linux aarch64
--   C lang standard:  11
--
-- BUILD_DEBIAN:
-- USE_BORINGSSL:          OFF
-- BORINGSSL_ROOT_DIR:
-- BORINGSSL_INCLUDE_DIR:
-- OPENSSL_VERSION:        3.5.6
-- OPENSSL_ROOT_DIR:
-- OPENSSL_INCLUDE_DIR:    /usr/include
-- OPENSSL_LIBRARIES:      /usr/lib64/libssl.so;/usr/lib64/libcrypto.so
-- PAM_LIBRARY:            /usr/lib64/libpam.so
-- PAM_INCLUDE_DIR:        /usr/include/security
-- BUILD_COMPRESSION:      TRUE
-- ZSTD_INCLUDE_DIR:       /usr/include
-- ZSTD_LIBRARY:           /usr/lib64/libzstd.so
-- ZLIB_INCLUDE_DIRS:      /usr/include
-- ZLIB_LIBRARIES:         /usr/lib64/libz.so
-- LDAP_SUPPORT:           TRUE
-- LDAP_LIBRARY:           /usr/lib64/libldap.so
-- LDAP_INCLUDE_DIR:       /usr/include
-- SYSTEMD_FOUND:          TRUE
-- SYSTEMD_LIBRARY:      /usr/lib64/libsystemd.so
-- SYSTEMD_INCLUDE_DIR:   /usr/include
-- USE_UCONTEXT:           OFF
-- MM_MEM_PROF:            OFF
-- USE_TCMALLOC:           OFF
[...]

PR #1465, opened 2026-07-17
fix machinarium writev build with compression enabled

mikhailnov:
Исправление ошибки сборки https://github.com/yandex/odyssey/issues/1462

Честно говоря, сделал полностью с помощью ИИ и сам НЕ вникал, мог получиться бред. Посмотрите, пожалуйста.

<hr>
machine_writev_raw() failed to compile under -DBUILD_COMPRESSION=True: 'iovec' was used instead of the 'iov' parameter, and mm_iov_advance() was referenced but never defined. The local-iov mutation would not have propagated to the caller anyway, so the partial-progress path for compression was dead on arrival.

Align the contract of machine_writev_raw() with the existing machine_write_raw() by adding a size_t *processed out-parameter. *processed is only written by the streaming compressor on partial-progress errors (see zstd_write/zlib_write in zpq_stream.c), which is exactly the contract machine_write_raw() already has via mm_zpq_write(). On success rc itself is the number of consumed input bytes uniformly across the socket, TLS and compression paths.

od_io_writev() advances the scatter-gather vector by processed + (rc > 0 ? rc : 0), matching the additive accounting already used by machine_write_no_free() for the machine_write_raw() path: on a retryable error with partial progress *processed holds the bytes the compressor has already accepted and they must not be resent on retry, otherwise the compressed stream gets corrupted.

Also make the iov argument of mm_compression_writev() const, since it only reads from it via mm_iovcpy(), so the const pointer from machine_writev_raw() can be forwarded without a cast.

Co-authored-by: [Z.AI](https://z.ai/subscribe?ic=OKXME28I1P) GLM

<hr>
Еще вот что ИИ выдал:

"od_io_writev выходит из цикла, как только iovcnt == 0, без проверки mm_compression_write_pending(io) — тогда как machine_write_no_free (write.c:99) такую проверку делает (while (total != size || mm_compression_write_pending(io))). Теоретически, если компрессор принял весь ввод, но ещё не сбросил внутренний TX-буфер, часть байт останется в компрессоре неотправленной. Поведение точно такое же, как в коде до коммита."

### DIFF
--- a/sources/include/machinarium/compression.h
+++ b/sources/include/machinarium/compression.h
@@ -17,7 +17,7 @@ static inline int mm_compression_is_active(mm_io_t *io)
 	return io->zpq_stream != NULL;
 }
 
-int mm_compression_writev(mm_io_t *io, struct iovec *iov, int n,
+int mm_compression_writev(mm_io_t *io, const struct iovec *iov, int n,
 			  size_t *processed);
 
 int mm_compression_read_pending(mm_io_t *io);
--- a/sources/include/machinarium/machinarium.h
+++ b/sources/include/machinarium/machinarium.h
@@ -252,7 +252,7 @@ MACHINE_API ssize_t machine_write_raw(mm_io_t *, const void *, size_t,
 				      size_t *);
 
 MACHINE_API ssize_t machine_writev_raw(mm_io_t *, const struct iovec *iov,
-				       int iovcnt);
+				       int iovcnt, size_t *processed);
 
 MACHINE_API int machine_write(mm_io_t *, machine_msg_t *, uint32_t time_ms);
 
--- a/sources/io.c
+++ b/sources/io.c
@@ -115,12 +115,27 @@ int od_io_writev(od_io_t *io, struct iovec *iov, int iovcnt,
 {
 	mm_io_set_deadline(io->io, timeout_ms);
 	while (iovcnt > 0) {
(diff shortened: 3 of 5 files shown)


krylosov-aa:
I'm not a maintainer, I just looked at the issue and this PR

`iovec` is not a variable here (the argument is `iov`), and `mm_iov_advance` does not exist. That leftover is from 6c7032b0, when `machine_writev_raw` stopped taking `machine_iov_t *`.

The zpq `processed` accounting in this PR matches `machine_write_raw`, but it is more than this bug needs. `od_io_writev` is only used for the backend (xplan), and listen `compression` is now parsed and ignored, so the writev+compression path does not run. Changing the `MACHINE_API` of `machine_writev_raw` and the long comments are not worth it. Format CI fails on those comments too, and the PR conflicts with current `io.c`.

I would just fix the names and drop the dead call:

```c
#ifdef MM_BUILD_COMPRESSION
    if (mm_compression_is_active(io)) {
        size_t processed = 0;
        rc = mm_compression_writev(io, iov, iov_to_write, &processed);
        (void)processed;
    } else if (mm_tls_is_active(io))
#else
    if (mm_tls_is_active(io))
#endif
```

Keep the dummy `processed` because `mm_zpq_write` writes it on a partial-progress error. Do not advance `iov` inside `machine_writev_raw` it is const, and the caller already advances on `rc > 0`.

PR #1466, opened 2026-07-17
OpenSSL 4 compatibility

blkerby:
I'm an Ubuntu Developer, and we're currently transitioning to OpenSSL 4. Currently, `odyssey` [fails to build](https://bugs.launchpad.net/ubuntu/+source/odyssey/+bug/2154864) with OpenSSL 4, with errors caused by the following:
- A const qualifier is added to the return type of `X509_get_subject_name` in OpenSSL 4, which needs to be reflected in the local variable's type.
- The function `X509_NAME_get_text_by_NID` is deprecated.

To avoid the deprecated function, this PR replaces uses `X509_NAME_get_index_by_NID`, `X509_NAME_get_entry`, and `X509_NAME_ENTRY_get_data`, following the guidance in the [manpage](https://linux.die.net/man/3/x509_name_get_text_by_nid). I confirmed that with this change, the `odyssey` package builds successfully in a [PPA](https://launchpad.net/~blkerby/+archive/ubuntu/openssl4-transition/+packages) with OpenSSL 4 dependencies.

### DIFF
--- a/sources/machinarium/tls.c
+++ b/sources/machinarium/tls.c
@@ -498,9 +498,12 @@ static inline int mm_tls_verify_name(char *cert_name, const char *name)
 int mm_tls_verify_common_name(mm_io_t *io, char *name)
 {
 	X509 *cert = NULL;
-	X509_NAME *subject_name = NULL;
-	char *common_name = NULL;
+	const X509_NAME *subject_name = NULL;
+	const X509_NAME_ENTRY *entry = NULL;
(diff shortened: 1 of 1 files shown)


rkhapov:
hi! thanks for the patch!

appended fix for compatibility with <= 4.0.0 versions
https://github.com/yandex/odyssey/pull/1466/changes/24295fe29cd4b68bea8a29c5e1248f748ef8e5b0

wll wait for CI and merge

rkhapov (approved):

PR #1475, opened 2026-07-23
try to fix duplex deadlock

rkhapov:


reshke (approved):

rkhapov:
need to check performance

PR #1507, opened 2026-08-08
introduce new counters for clients waiting for a route

evkuzin:
SHOW POOLS reported cl_active and cl_waiting but never count_queue, leaving clients blocked on an exhausted pool invisible to monitoring, while maxwait was hardcoded to 0. Add a cl_queue column plus a real maxwait derived from the oldest client in the queue list, and export both through the Prometheus exporter as odyssey_client_pool_queue_route.
Also make exporter more friendly for new fields in the views.


evkuzin:
Nailed it from the first time! All green 🥇

sources/router.c
```diff
@@ -945,6 +945,14 @@ od_router_try_attach(od_router_t *router, od_client_t *client,
 
 	od_route_lock(route);
 
+	/*
+	 * stamp on the first transition only: the caller retries in a loop
+	 * and re-stamping would keep resetting the measured waiting time
+	 */
+	if (client->state != OD_CLIENT_QUEUE) {
+		client->time_queue_start = machine_time_us();
```
rkhapov, line 953: there is already wait time measuring at od_router_attach, why need adding another one?

evkuzin: I think the other one is stack local for waiting coroutine, no?

evkuzin: also it seems like its computed after wait finished, while this reports while we are waiting.

rkhapov (approved):

PR #1528, opened 2026-08-20
fix broken tls reload if cert path changed

evkuzin:
**fix broken tls reload if cert path changed**

SIGHUP never applied a changed tls_cert_file / tls_key_file / tls_ca_file. The reload built the new TLS handle from server->config (the old listen config) so only tls_mode was ever picked up. A pooler whose cert path changed kept serving the old certificate until a full restart, eventually serving an expired one.

Fixed by building the handle from the freshly parsed listen_config. The paths are copied into the handle, so they stay valid after the parsed config is freed.

**reject a broken tls certificate instead of serving none**

Certificate files are only opened during a handshake, so a nonexistent or mismatched cert passed odyssey --test ("config is valid") and passed startup — then failed every TLS connection, with nothing logged. Certificates are now loaded during od_config_validate:

- require / verify_ca / verify_full — refuse to start; TLS is mandatory and cannot work
- allow / prefer — start and log the error; the listener still serves plaintext clients
- on reload — keep the working certificate rather than replacing it with a broken one

**report a failed config load in SHOW LISTS EXTENDED**

When a reload fails, odyssey keeps running on its previous in-memory config which is correct, but it leaves no indication that the file on disk is unusable and a restart would not come up. SHOW LISTS now reports config_load_failed, exported as odyssey_config_load_failed, covering all four reload failure paths (od_cfg_import, od_config_validate, od_rules_validate, od_rules_autogenerate_defaults).

### DIFF
--- a/sources/console.c
+++ b/sources/console.c
@@ -40,6 +40,7 @@ typedef enum {
 	OD_LPREPARED_STMTS,
 	OD_LCLIENTS,
 	OD_LLISTS,
+	OD_LLISTS_EXTENDED,
 	OD_LHELP,
 	OD_LSET,
 	OD_LCREATE,
@@ -76,6 +77,7 @@ static od_keyword_t od_console_keywords[] = {
 	od_keyword("global_prepared_statements", OD_LPREPARED_STMTS),
 	od_keyword("clients", OD_LCLIENTS),
 	od_keyword("lists", OD_LLISTS),
+	od_keyword("lists_extended", OD_LLISTS_EXTENDED),
 	od_keyword("set", OD_LSET),
 	od_keyword("pools", OD_LPOOLS),
 	od_keyword("pools_extended", OD_LPOOLS_EXTENDED),
@@ -270,7 +272,7 @@ static inline int od_console_show_help(machine_msg_t *stream)
 		"\n"
 		"Console usage\n"
 		"\tSHOW STATS|HELP|POOLS|POOLS_EXTENDED|DATABASES|SERVER_PREP_STMTS|SERVERS|CLIENTS|HOST_UTILIZATION\n"
-		"\tSHOW LISTS|ERRORS|ERRORS_PER_ROUTE|VERSION|VERSION_EXTENDED|LISTEN|STORAGES\n"
+		"\tSHOW LISTS|LISTS_EXTENDED|ERRORS|ERRORS_PER_ROUTE|VERSION|VERSION_EXTENDED|LISTEN|STORAGES\n"
 		"\tSHOW CONFIG|RULES|FDS|IS_PAUSED|GLOBAL_PREPARED_STATEMENTS\n"
 		"\tKILL_CLIENT <client_id>\n"
 		"\tRELOAD\n"
@@ -2001,8 +2003,9 @@ static inline int od_console_show_lists_cb(od_route_t *route, void **argv)
 	return 0;
 }
 
-static inline int od_console_show_lists(od_client_t *client,
-					machine_msg_t *stream)
+static inline int od_console_show_lists_impl(od_client_t *client,
+					     machine_msg_t *stream,
+					     int extended)
 {
 	od_assert(stream);
 	od_router_t *router = client->global->router;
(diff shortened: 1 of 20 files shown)


evkuzin:
Hm... clearly I did something wrong. In the tests

evkuzin:
Hm... I actually think this leak was there even before me. 
I'll work on that

sources/system.c
```diff
@@ -563,33 +595,63 @@ void od_system_config_reload(od_system_t *system)
...
+		int files_changed =
+			listen_config != NULL &&
+			!od_tls_opts_files_eq(server->config->tls_opts,
+					      listen_config->tls_opts);
+
+		machine_tls_t *tls = od_tls_frontend(tls_source);
+		if (tls == NULL) {
+			od_error(
+				&instance->logger, "reload-config", NULL, NULL,
+				"failed to build tls handler for %s:%d, keeping previous certificate",
```
rkhapov, line 695: i am not sure if it is acceptable..

tls certs is about security, and the approach of 'do your best in certs reloading or just fail' maybe is not enough for the case (yes, current impl seems questinable too)

maybe it is more safe to do smth like exit(1) ?

evkuzin: Ugh. You mean exit odyssey? 
In the current situation we will continue serving possibly still valid (for some time) cert. Maybe a message should be just a little be more like
"Aaaaaaaa, oh noooooo, we cannot reload cert, it will soon expire!!111 something is wroooooong!!!111"

evkuzin: Added a new logic. for broken tls etc... check whole PR again pls.

evkuzin: Also I think postgres does same thing https://github.com/postgres/postgres/commit/de41869b64d57160f58852eab20a27f248188135

evkuzin:
that is odd. It failed only exporter test and now 3 more when I fixed it, maybe they are flaky. I'll retrigger...

sources/system.c
```diff
@@ -504,25 +552,19 @@ void od_system_config_reload(od_system_t *system)
 	rc = od_cfg_import(&instance->logger, &config, &rules, system->global,
 			   &hba_rules, instance->config_file);
 	if (rc == -1) {
-		od_rules_unlock(&router->rules);
-		od_config_free(&config);
-		od_rules_free(&rules);
+		od_system_config_reload_abort(system, &config, &rules);
```
rkhapov: it is better not to do it with separated function, it makes code harder to read..

for C project either choose the wrapper-function or goto error style

sources/console.c
```diff
@@ -2023,6 +2024,13 @@ static inline int od_console_show_lists(od_client_t *client,
 	if (rc == NOT_OK_RESPONSE) {
 		return NOT_OK_RESPONSE;
 	}
+	/* config_load_failed */
+	rc = od_console_show_lists_add(
```
rkhapov, line 2126: adding new field to show lists command will break the backward compatibility.. lets add show lists_extended with all field from show lists and new field

sources/system.c
```diff
@@ -218,6 +218,27 @@ od_system_server_t *od_system_server_init(void)
...
 }
 
+static inline od_retcode_t
+od_system_server_retire_tls(od_system_server_t *server)
+{
+	if (server->tls == NULL) {
+		return OK_RESPONSE;
+	}
+
+	machine_tls_t **retired = od_realloc(
```
rkhapov: not so important, but mm_vector_t can be used here

sources/system.c
```diff
@@ -623,6 +705,9 @@ void od_system_config_reload(od_system_t *system)
 	       "%d routes created/deleted and scheduled for removal", updates);
 
 	od_rules_groups_checkers_run(&instance->logger, &router->rules);
+
+	/* the file on disk loaded, so a restart would come up on it */
+	atomic_store(&instance->config_load_failed, 0);
```
rkhapov, line 754: maybe make it add-only? for now the situations of one and two failed allocation has no difference

sources/system.c
evkuzin: Thx! Will do.

sources/console.c
evkuzin: Right, I forgot we did same for version. Going to fix it.

sources/system.c
evkuzin: I missed this one, will look into it.

sources/system.c
evkuzin: What would be the benefit? Currently its written as a switch. config either in a healthy state or broken. If it is broken - there might be 4 at this time possible conditions, and we don't elaborate on what is broken, we just tell that it is and the human work is to go to the logs and check what is broken, fix, reload again and see if it flipped to 0. With switching it to a counter we don't get much of benefit here, do we? Or maybe I'm missing something?

rkhapov (approved):

PR #1531, opened 2026-08-20
docs: state that the LDAP search scope is not configurable

### LINKED ISSUE #1529: ldapscope is accepted by the config parser but the search scope is not configurable
## Summary

The `ldap_endpoint` section accepts an `ldapscope` option, but the search
scope is not configurable: the only directory search in the tree passes the
constant `LDAP_SCOPE_SUBTREE`. The value is parsed, stored and printed back
when the configuration is dumped, so nothing tells the administrator that it
is not taken into account.

The documentation does not mention `ldapscope` at all, so there is currently
no way to find out other than by reading the source.

## Reproduction

At commit `82243bde`, every occurrence of `ldapscope` outside the generated
parser:

```bash
git grep -n "ldapscope" -- sources/ | grep -v "cfg/parse.y"
```

```
sources/cfg/convert.c:1702:             COPY_LDAP_STR(ldapscope);
sources/cfg/keywords.c:127:     { "ldapscope", LDAP_SCOPE },
sources/cfg/model.c:160:        dump_string(file, "ldapscope", 1, &e->ldapscope);
sources/cfg/model.c:664:        od_cfg_string_field_free(&endpoint->ldapscope);
sources/include/cfg/model.h:251:        od_cfg_string_field_t ldapscope;
sources/include/ldap_endpoint.h:25:     char *ldapscope;
sources/ldap.c:142:     le->ldapscope = NULL;
sources/ldap.c:192:     if (le->ldapscope) {
sources/ldap.c:193:             od_free(le->ldapscope);
```

Every line registers, copies, dumps or frees the value. None reads it.

The search itself is the only call to `ldap_search_s` in the tree, and its
scope argument is a literal:

```bash
git grep -n "ldap_search_s\|LDAP_SCOPE_" -- sources/ | grep -v "cfg/"
```

```
sources/include/od_ldap.h:15:/* For functions ldap_unbind, ldap_search_s, ldap_simple_bind_s */
sources/ldap.c:495:     int rc = ldap_search_s(ldap, endp->ldapbasedn, LDAP_SCOPE_SUBTREE,
```

This is long-standing behaviour rather than a regression: the option was
introduced together with LDAP authentication in `74623edf` ("LDAP auth in
od", April 2021), and the same commit already registered the keyword and the
field without ever passing it to the search.

## Expected behavior

The documentation should state what Odyssey actually does — that the search
always covers the subtree below `ldapbasedn`, and that `ldapscope` does not
change it. This is the same behaviour as PostgreSQL, which has no separate
search scope option either, so the current implementation matches the system
Odyssey sits in front of.

The `ldap_endpoint` section of `docs/configuration/rules.md` is the natural
place: it already explains how the client DN is determined, and this is one
short paragraph next to it. Documentation only, no source or behaviour
changes. I will open a pull request with it.

## Scope

In scope: `docs/configuration/rules.md`.

Out of scope, mentioned only for reference: whether the option should
eventually become effective, or be dropped from the grammar. Either would be
a behaviour change and is a separate discussion; documenting what happens
today is correct regardless of how that discussion goes.

## Environment

Odyssey at commit `82243bde` (branch `master`), Ubuntu 22.04 (WSL2), x86_64.
Findings are from reading the tree; no local modifications
(`git diff --stat` is empty).

YdiBik:
Closes #1529

## What changed

Added one paragraph to the `ldap_endpoint` section of
`docs/configuration/rules.md`, stating that the search scope is not
configurable and that `ldapscope` is accepted but not used.

Documentation only. No source changes, no behaviour changes.

## Why

`ldapscope` is parsed, stored in the configuration model and printed back
when the configuration is dumped, so a value set there looks like it took
effect. The search always passes the constant `LDAP_SCOPE_SUBTREE`
(`sources/ldap.c:495`, the only `ldap_search_s` call in the tree), and the
documentation did not mention the option at all.

The paragraph describes the current behaviour and does not prejudge whether
the option should later become effective or be removed.

## Verification

- `sources/ldap.c:495` is the only call to `ldap_search_s`; its scope
  argument is the literal `LDAP_SCOPE_SUBTREE`.
- All other occurrences of `ldapscope` (`sources/cfg/convert.c:1702`,
  `keywords.c:127`, `model.c:160` and `:664`, `include/cfg/model.h:251`,
  `include/ldap_endpoint.h:25`, `ldap.c:142`, `:192`) register, copy, dump or
  free the value; none reads it.
- The statement about PostgreSQL follows its documentation of LDAP
  authentication, which describes the search+bind search as covering the
  subtree at `ldapbasedn` and offers no separate scope option.

`make format` was not run: the change touches no C source, and `clang-format`
does not apply to Markdown. The added paragraph follows the wording and
layout of the surrounding text in the same section.

## Environment

Ubuntu 22.04 (WSL2), x86_64. Branched from `upstream/master` at `82243bde`.

### DIFF
--- a/docs/configuration/rules.md
+++ b/docs/configuration/rules.md
@@ -313,6 +313,12 @@ The way the client DN is determined depends on whether `ldapbasedn` is set:
   performed: the client DN is formed by concatenating `ldapprefix`, the user
   name the client connected with, and `ldapsuffix`.
 
+The scope of the search in search+bind mode is not configurable: Odyssey
+always searches the whole subtree below `ldapbasedn`. The `ldapscope` option
+is accepted by the configuration parser and shown when the configuration is
+dumped, but its value is not used. PostgreSQL behaves the same way and has no
+separate option for the search scope either.
+
 In both modes Odyssey first opens the connection to the LDAP server and binds
 with `ldapbinddn` / `ldapbindpasswd`, or anonymously when they are not set.
 `ldapprefix` and `ldapsuffix` are ignored when `ldapbasedn` is set. The names


rkhapov:
thanks!

rkhapov (approved):

PR #1533, opened 2026-08-21
introduce show config command

### LINKED ISSUE #1532: SHOW CONFIG: no way to inspect the configuration a running instance is using
### Problem

There is no way to ask a running Odyssey instance which configuration it is
actually using.

`show config` is not recognised by the admin console — the keyword is absent
from `od_console_keywords[]` in `sources/console.c`:

```
console=> show config;
ERROR:  odyssey: cd126761dc015: console command error: show config
```

Reading `odyssey.conf` is not equivalent, because the file and the running
process can disagree. `od_config_reload()` (`sources/config.c:111`) copies
twenty fields from the freshly parsed configuration into `instance->config`:

`log_debug`, `log_config`, `log_session`, `log_query`, `log_stats`,
`stats_interval`, `client_max_set`, `client_max`, `client_max_routing`,
`max_sigterms_to_die`, `server_login_retry`, `backend_connect_timeout_ms`,
`cancel_timeout_ms`, `smart_search_path_enquoting`, `disable_nolinger`,
`keepalive`, `keepalive_keep_interval`, `keepalive_probes`,
`keepalive_usr_timeout`, `graceful_shutdown_timeout_ms`.

Every other field of `struct od_config` keeps the value the process started
with — `workers`, `resolvers`, `readahead`, `nodelay`, `dns_cache_ttl`,
`virtual_processing`, `group_checker_interval` and the rest. `RELOAD` returns
success and logs nothing about it, so an operator who edits the file and
reloads has no way to tell which half of the change took effect.

The console already exposes `SHOW RULES`, `SHOW LISTEN`, `SHOW STORAGES` and
`SHOW STATS`, but nothing that reports global configuration values.

### Previous discussion

This was requested in #210 (September 2020), with the same reproduction and
the same comparison to PgBouncer. The request was welcomed, with one
condition — credentials must be filtered out. That issue was closed in May
2021 together with `2019c641` ("odyssey configuration from CLI"), which added
command-line option parsing and did not touch `sources/console.c`, so the
console command itself was never implemented.

### Why it matters

Odyssey is frequently adopted by teams moving from PgBouncer, where
`SHOW CONFIG` is part of everyday operation: comparing two poolers behind a
load balancer, verifying a reload, feeding dashboards and monitoring
templates.

### Proposal

Add `SHOW CONFIG` to the admin console, reporting for every global parameter:
its name as accepted by the configuration parser, the value the process is
using, the built-in default, and whether `RELOAD` applies a new value. The
column set matches PgBouncer's `SHOW CONFIG` — `key`, `value`, `default`,
`changeable` — so existing tooling sees a familiar shape.

On the credentials condition from #210: `od_config_t` holds no secrets.
Passwords, SCRAM secrets and LDAP bind credentials live in the rules
(`od_rule_t`), which this command does not touch — those are reported by
`SHOW RULES`.

I have an implementation with a functional test. PR follows.

YdiBik:
Closes #1532

Adds `SHOW CONFIG` to the admin console. For every global configuration
parameter it reports the name accepted by the configuration parser, the value
the running process is using, the built-in default, and whether `RELOAD`
applies a new value without a restart.

```
console=> show config;
             key             | value | default | changeable
-----------------------------+-------+---------+------------
 workers                     | 2     | 1       | no
 log_query                   | yes   | no      | yes
 ...
(57 rows)
```

The column set — `key`, `value`, `default`, `changeable` — matches
PgBouncer's `SHOW CONFIG`, so tooling written against it sees a familiar
shape.

### Why the command is not the same as reading odyssey.conf

`od_config_reload()` copies twenty fields into the running instance; the rest
keep the value the process started with, and `RELOAD` reports success either
way. The `changeable` column makes the difference visible, so comparing the
output against the file is enough to spot a parameter that needs a restart.

### Implementation notes

`od_config_fields[]` in `sources/config.c` describes each parameter by name,
type, offset inside `od_config_t` and reloadability. The console walks the
table and stays unaware of how the configuration is laid out.

Defaults are not stored in the table. `od_config_field_default()` initialises
a throwaway `od_config_t` through `od_config_init()` and reads the same
offset, so the reported defaults cannot drift away from the real ones, and
computed defaults such as `readahead` are reported correctly.

The `reloadable` flag has no such source and is kept in sync with
`od_config_reload()` by hand; there is a comment above the table saying so.

`od_config_print()` is left untouched. It could be rebuilt on the same table
later, but that would change the format of an existing log output, which
seems better kept out of this change.

### Credentials

`od_config_t` holds no secrets — passwords, SCRAM secrets and LDAP bind
credentials live in `od_rule_t` and are reported by `SHOW RULES`. This
addresses the condition raised in #210.

### Scope

Global scalar parameters only, 57 of them. Not included: the `listen`
section (`SHOW LISTEN`), the nested `soft_oom`, `query_parsing`,
`conn_drop_options` and `cpu_affinity` structures, and per-route settings
(`SHOW RULES`). Those are not key/value pairs and would not fit the same
output.

### Other changes

`SHOW HELP` did not list `RULES`, `FDS`, `IS_PAUSED`, `VERSION_EXTENDED`,
`GLOBAL_PREPARED_STATEMENTS`, `PAUSE` or `RESUME`, all of which the
dispatcher handles. Adding `CONFIG` to the same text made that obvious, so
the list is brought in line with the code. Happy to split this out if
preferred.

### Testing

`test/functional/tests/show-config/` covers three cases:

- the file changes and no `RELOAD` is issued — the output does not follow;
- `RELOAD` applies a reloadable parameter — the output changes;
- `RELOAD` does not apply `workers` — the output keeps the old value and
  `changeable` says `no`.

Plus output shape: four columns on every row, `changeable` is `yes` or `no`,
no empty values, defaults independent of the file.

Values longer than 255 characters are truncated; this is documented.

`docs/features/console.md` describes the command, the columns, the truncation
limit, and the one parameter whose default is derived from another
(`client_max_routing` becomes `workers * 64` when unset).

### DIFF
--- a/sources/config.c
+++ b/sources/config.c
@@ -17,6 +17,7 @@
 #include <router.h>
 #include <config.h>
 #include <od_memory.h>
+#include <util.h>
 
 void od_config_init(od_config_t *config)
 {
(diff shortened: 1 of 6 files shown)


sources/config.c
```diff
@@ -336,6 +339,196 @@ static inline char *od_config_yes_no(int value)
...
+void od_config_field_value(od_config_t *config, const od_config_field_t *field,
+			   char *buf, size_t size)
+{
+	od_config_field_format(config, field, buf, size);
+}
+
+void od_config_field_default(const od_config_field_t *field, char *buf,
+			     size_t size)
+{
+	od_config_t defaults;
```
rkhapov: od_config_t is a large object.. this can lead to stack overflow on the coroutines

AdamerGitHub:
The Functional (build_dbg, clang) job failed in /tests/prep_stmts with `could not open file "base/16388/16477": Interrupted system call`, before reaching the test added here. The other five functional jobs passed, and the previous run on this branch passed as well. The same matrix cell failed on an unrelated PR too (run 32574811153, in /tests/npgsql_compat), so this looks like flakiness in that configuration rather than something introduced here.

sources/config.c
AdamerGitHub: Good point, fixed. The defaults are now built once per process into a static instance guarded by pthread_once, so nothing of that size lands on a coroutine stack, and od_config_init() is no longer called once per parameter.

sources/config.c
```diff
@@ -337,6 +337,195 @@ static inline char *od_config_yes_no(int value)
...
+	if (index >= od_config_fields_count()) {
+		return NULL;
+	}
+	return &od_config_fields[index];
+}
+
+void od_config_field_value(od_config_t *config, const od_config_field_t *field,
+			   char *buf, size_t size)
+{
+	char *at = (char *)config + field->offset;
```
rkhapov, line 483: minor, but int8_t/uint8_t would be more portable

rkhapov (approved):

PR #1586, opened 2026-09-10
Fix application_name reset in session pooling

### LINKED ISSUE #1516: `RESET` and `DISCARD ALL` lose startup `application_name` in session pooling
**Odyssey version**

`1.5.3-rc1` (`1d082b6`, session pool mode)

**Description**

With `maintain_params=yes` (the default), a connection started with `application_name=od_reset_probe` initially reports that value. In Odyssey session pooling, `RESET application_name`, `RESET ALL`, and `DISCARD ALL` change it to `''`; direct PostgreSQL restores `od_reset_probe`.

The physical backend StartupMessage does not include the frontend's `application_name`, so PostgreSQL initializes it as `''` [`backend.c`](https://github.com/yandex/odyssey/blob/1d082b6fc601035eaeac484bd0797a7f64e75399/sources/backend.c#L340-L385). `od_deploy()` later synchronizes `client->vars` to that backend [`deploy.c`](https://github.com/yandex/odyssey/blob/1d082b6fc601035eaeac484bd0797a7f64e75399/sources/deploy.c#L43-L64). When PostgreSQL reports the reset value, `od_backend_update_parameter()` copies `''` into both `server->vars` and `client->vars`, replacing the frontend startup value [`backend.c`](https://github.com/yandex/odyssey/blob/1d082b6fc601035eaeac484bd0797a7f64e75399/sources/backend.c#L1068-L1078).

**Reproduce**

1. Start PostgreSQL 16 and Odyssey with `pool "session"`, leaving `maintain_params` at its default of `yes`.
2. Run the following with the direct and Odyssey DSNs supplied by the environment:

```python
#!/usr/bin/env python3
import psycopg


APP = "od_reset_probe"
ACTIONS = ("RESET application_name", "RESET ALL", "DISCARD ALL")


def run(label: str, dsn: str) -> None:
    try:
        for action in ACTIONS:
            with psycopg.connect(dsn, application_name=APP, autocommit=True) as conn:
                before = conn.execute("SHOW application_name").fetchone()[0]
                conn.execute(action)
                after = conn.execute("SHOW application_name").fetchone()[0]
            print(f"{label}: {action}: {before!r} -> {after!r}")
    except Exception as exc:
        print(f"{label}: {type(exc).__name__}: {exc}")


run("direct", DIRECT_DSN)
run("odyssey", ODYSSEY_DSN)
```

**Expected behavior**

Each command should restore the `application_name` supplied by that frontend's startup packet, matching direct PostgreSQL.

**Actual behavior**

Direct PostgreSQL returns:

```text
direct: RESET application_name: 'od_reset_probe' -> 'od_reset_probe'
direct: RESET ALL: 'od_reset_probe' -> 'od_reset_probe'
direct: DISCARD ALL: 'od_reset_probe' -> 'od_reset_probe'
```

Odyssey returns:

```text
odyssey: RESET application_name: 'od_reset_probe' -> ''
odyssey: RESET ALL: 'od_reset_probe' -> ''
odyssey: DISCARD ALL: 'od_reset_probe' -> ''
```

krylosov-aa:
In session pooling with `maintain_params yes`, `RESET application_name`, `RESET ALL` and `DISCARD ALL` reset `application_name` to an empty string instead of its initial value. Odyssey applies the client value through `SET`, which does not change PostgreSQL's reset value.

Pass the client's initial `application_name` in the backend startup message. Reconnect a pooled backend when the next client's initial name differs. Clients with the same initial name can reuse the connection.

Fixes #1516

### DIFF
--- a/sources/backend.c
+++ b/sources/backend.c
@@ -338,7 +338,7 @@ int od_backend_startup(od_server_t *server, kiwi_params_t *route_params,
 
 #define DEFAULT_ARGV_SIZE 6
 
-	kiwi_fe_arg_t argv[DEFAULT_ARGV_SIZE +
+	kiwi_fe_arg_t argv[DEFAULT_ARGV_SIZE + 2 +
 			   2 * route->rule->backend_startup_vars_sz];
 
 	kiwi_fe_arg_t default_argv[] = {
@@ -380,6 +380,22 @@ int od_backend_startup(od_server_t *server, kiwi_params_t *route_params,
 		argv[argc + 1].name = "database";
 		argv[argc + 1].len = 9;
 		argc += 2;
+	}
+
+	kiwi_var_unset(&server->startup_application_name);
+	if (route_params == NULL && client != NULL &&
+	    client->type == OD_POOL_CLIENT_EXTERNAL &&
+	    client->rule->maintain_params &&
+	    client->rule->pool->pool_type == OD_RULE_POOL_SESSION) {
+		kiwi_var_t *app = &client->startup_application_name;
+		if (app->value_len > 0) {
+			argv[argc].name = app->name;
+			argv[argc].len = app->name_len;
+			argv[argc + 1].name = app->value;
+			argv[argc + 1].len = app->value_len;
+			argc += 2;
+			server->startup_application_name = *app;
+		}
 	}
 
 	machine_msg_t *msg;
--- a/sources/client.c
+++ b/sources/client.c
@@ -39,6 +39,8 @@ void od_client_init(od_client_t *client)
 
 	kiwi_be_startup_init(&client->startup);
 	kiwi_vars_init(&client->vars);
+	kiwi_var_init(&client->startup_application_name, "application_name",
+		      sizeof("application_name"));
 	kiwi_key_init(&client->key);
 
 	od_io_init(&client->io);
--- a/sources/frontend.c
+++ b/sources/frontend.c
@@ -520,6 +520,18 @@ static inline od_frontend_status_t od_frontend_attach_to_endpoint(
 				continue;
 			}
 		}
+		if (!od_backend_need_startup(server) && route_params == NULL &&
+		    client->rule->maintain_params &&
+		    client->rule->pool->pool_type == OD_RULE_POOL_SESSION &&
+		    !kiwi_var_compare(&client->startup_application_name,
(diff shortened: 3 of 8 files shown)


krylosov-aa:
@rkhapov

rkhapov:
> Reconnect a pooled backend when the next client's initial name differs. 

this seems very strange idea for connection pooling

krylosov-aa:
The reconnect was intended to preserve the next client's reset value. PostgreSQL treats `application_name` passed at startup as the session default; a later `SET` changes the current value but not that default ([startup](https://www.postgresql.org/docs/18/protocol-flow.html#PROTOCOL-FLOW-START-UP), [RESET](https://www.postgresql.org/docs/18/sql-reset.html)).

For example, if a backend starts with `app_a` and is later reused by a client with `app_b`, `SET application_name = 'app_b'` updates the current name, but `RESET application_name` still restores `app_a`. Reconnecting avoids that, but I agree that losing backend reuse across different names is a significant drawback.

An alternative would be to store each client's initial name in Odyssey and handle explicit `RESET application_name`, `RESET ALL` and `DISCARD ALL` commands there. This would need to preserve transaction semantics and command ordering in both query protocols. Handling only client-issued commands would leave resets inside server-side functions using the backend's own default.

Where do we want to take this issue? Or should we not do anything about it?

rkhapov:
i think the better way would be recheck, if we can set application name in od_deploy..

krylosov-aa:
`od_deploy()` sets `application_name` through `kiwi_vars_cas()`, which generates `SET application_name = ...` when the client and backend values differ.

On PostgreSQL 18.4:


|  | `setting` | `reset_val` |
| ------------------------------ | --------- | ----------- |
| Backend startup with `app_a`   | `app_a`   | `app_a` |
| `SET application_name='app_b'` | `app_b`   | `app_a` |
| `RESET application_name`       | `app_a`   | `app_a` |

`RESET ALL` and `DISCARD ALL` also restored `app_a`. Using `set_config(..., false)` instead of `SET` left `reset_val` unchanged as well.

So setting the current name in `od_deploy()` works, but it does not change what subsequent resets restore.

Would it be enough to handle `RESET application_name`, `RESET ALL` and `DISCARD ALL` sent by the client? Or should those work the same as a direct PostgreSQL connection?

PR #1590, opened 2026-09-11
Use OpenSSL descriptions for certificate verification errors

### LINKED ISSUE #114: More debugging information on SSL_get_verify_result() error
I am getting `SSL_get_verify_result()` when attempting to connect to a TLS-enabled odyssey server (i.e. under the `listen` directive). However, I am failing to understand what is actually being verified on the peer certificate and where should I be looking at.

I get the impression pgbouncer is less strict on certificate validation and therefore is allowing me to continue with a potentially 'bad' certificate, but without more information it's difficult to understand what's the root cause.

Is there anything odyssey can call on openssl beforehand that it can print in order to avoid this rather generic error?

krylosov-aa:
Certificate verification errors missing from the hardcoded table were reported as `unknown`. Use `X509_verify_cert_error_string()` and include the numeric verification code, for example: `SSL_get_verify_result(): 47 (permitted subtree violation)`.

Fixes #114


krylosov-aa:
@rkhapov

rkhapov (approved):

PR #1592, opened 2026-09-12
Escape client fields in text logs

krylosov-aa:
Newlines and tabs in client values can split log records and TSKV fields. Apply the existing `%M` escaping rules to `%u`, `%d`, `%a`, and `%x`. The original client values are preserved; `%m` retains its raw output.

Reserve space for the final newline and the async buffer's NUL terminator when truncating text logs. Treat application_name without a stored value as missing.

### DIFF
--- a/docs/configuration/global.md
+++ b/docs/configuration/global.md
@@ -175,6 +175,7 @@ Supported flags:
 %u = user name
 %d = database name
 %a = application_name
+%x = external user ID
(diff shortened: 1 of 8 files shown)


krylosov-aa:
@rkhapov

krylosov-aa:
@rkhapov Could you please take a look at this PR and the others?

PR #1602, opened 2026-09-16
router: optimize cancel lookup with client hashmap

OpiumCat:
replace O(routes x servers) cancel search with O(1) lookup via a global client hashmap keyed by kiwi_key_t.
drop server refcounting; lock on the client instead.
frontend cleanup is simplified: all resources are freed once at the end instead of on each error path.


sources/router.c
```diff
@@ -926,11 +934,14 @@ static inline od_router_status_t
...
 		     bool wait_for_idle, od_storage_endpoint_t *endpoint)
 {
+	od_instance_t *instance = router->global->instance;
 	od_server_t *server;
 	od_route_t *route = client->route;
 
 	uint64_t now_us = machine_time_us();
 
+	mm_hashmap_keylock_t client_klock;
+	od_instance_clients_lock(instance, &client->key, &client_klock);
```
rkhapov: do we need to get lock on attach? seems like there will be cancel race anaway, but we will still lose some points in performance?

sources/system.c
```diff
@@ -831,6 +831,12 @@ static inline void od_system(void *arg)
 		return;
 	}
 
+	instance->clients = mm_hashmap_create(65536, 16, sizeof(kiwi_key_t),
```
rkhapov: need to think twice about constants..

1) the maximum clients amount i met was smth about 30-40k, while getting some rare collision will not hurt much, maybe lower the bucket count?

2) seems like we can approximate locks count with workers count?

sources/frontend.c
```diff
@@ -2871,9 +2872,10 @@ static void od_application_name_add_host(od_client_t *client)
 		      app_name_with_host, length + 1); /* return code ignored */
 }
 
-void od_frontend(void *arg)
+void od_frontend_implementation(od_client_t *client, bool *io_attached,
+				bool *routing_slot_released, bool *routed_ok,
```
rkhapov: it is hard to tell, that this approach is easier to support and understand, than the way is was before.. maybe separate this change into different PR?

sources/include/server.h
```diff
@@ -148,6 +129,8 @@ static inline void od_server_init(od_server_t *server, int reserve_prep_stmts)
 	}
 }
 
+void od_server_attach_client_locked(od_server_t *server, od_client_t *client);
+void od_server_detach_client_locked(od_server_t *server);
 void od_server_attach_client(od_server_t *server, od_client_t *client);
```
rkhapov, line 132: why not doing locks magic inside already existed od_server_attach_client/od_server_detach_client ?

sources/include/instance.h
```diff
@@ -48,3 +53,14 @@ void od_instance_set_shutdown_worker_id(od_instance_t *instance, int64_t id);
 int64_t od_instance_get_shutdown_worker_id(od_instance_t *instance);
 
 od_global_pstmt_map_t *od_instance_get_pstmts_map(od_instance_t *instance);
+
+mm_hash_t od_instance_clients_hm_hash(const void *a);
```
rkhapov: hash and cmp function seems like impl-private, no need to declare it in header file

sources/include/instance.h
```diff
@@ -35,6 +39,7 @@ struct od_instance {
 	} cmdline;
 
 	od_global_pstmt_map_t *pstmts;
+	mm_hashmap_t *clients;
```
rkhapov: clientsById? or even maybe idToClient ?

rkhapov: the best approach will be creating new type, which will encapsulate the hashmap field i think..

rkhapov:
in general: this is very good idea

but we also need to ensure there is no performance degradation in high pool connections usage: locks on hot path when transaction pooling enabled can make a difference..

OpiumCat:
1. Frontend refactor will be done in a separate PR

2. No latency regression on the hot path. Benchmarked against the previous cancel lookup:
- End-to-end cancel latency (10 clients): ~3.0 ms both before and after — difference is within noise
- Hot-path pgbench (-C, connect-per-transaction, 20 clients): ~8556 tps vs ~8400 tps — the new code is slightly faster (~1.9%), no regression

3. Hashmap memory. The map is created once at startup with mm_hashmap_create(32768, 16 × workers, sizeof(kiwi_key_t), sizeof(od_client_t*), ...).
- Fixed cost (independent of client count): ~877 KB theoretical — dominated by the bucket array (36,353 buckets × 24 B = 872 KB) plus 71 lock mutexes (64 B each). Measured RSS delta vs the old build is ~608 KB (less than theoretical).
- Per-client cost: 40 B per entry (24 B kvp header + 8 B key + 8 B value). Negligible: 1000 clients ≈ 39 KB, 10000 clients ≈ 391 KB.
- Overall the map adds ~1.5% to odyssey's ~41 MB baseline RSS.

PR #1612, opened 2026-09-20
Keep group checkers running when config reload is rejected

### LINKED ISSUE #1611: Failed configuration reload stops group membership checks
When Odyssey rejects a new configuration, it logs that it is keeping the running configuration. However, existing group checkers stop and their member lists are cleared. Users who could connect through a group rule can no longer match that rule.

With a default rule using `authentication "block"`, new connections from these users fail with `FATAL: user blocked`. Changes to group membership are no longer picked up.

Steps to reproduce:

1. Create a table:
   ```sql
   CREATE TABLE group_members (name text);
   INSERT INTO group_members VALUES ('alice');
   ```
2. Configure an Odyssey group rule with `group_query "SELECT name FROM group_members"` and a working internal route for `group_query_user` / `group_query_db`. Use `authentication "block"` for the default user rule and `group_checker_interval 100`.
3. Wait until a new connection as `alice` can run `SELECT 1`. Check that `bob` is blocked.
4. Add `invalid_reload_option yes` to the Odyssey configuration and reload it.
5. Check that `SHOW INSTANCE` reports `config_load_failed=1`.
6. Try a new connection as `alice`: it is now blocked.
7. Insert `bob` into `group_members`: connections as `bob` remain blocked after several checker intervals.

The problem occurs with both SQL `RELOAD` and SIGHUP. PostgreSQL remains available. Restoring a valid configuration and reloading it restores group access.

An unsuccessful reload should leave existing group rules working, including continued detection of membership changes.

krylosov-aa:
A config reload stops group checkers before reading the new configuration. If the file is missing or invalid, Odyssey keeps the old rules, but their group membership has already been cleared. Existing group members can lose access on new connections, and later membership changes are no longer detected.

This change completes configuration validation, default rule generation and sorting before stopping the running checkers. It also checks sorting errors and frees temporary groups, storages and HBA rules when preparation fails. Rejected reloads now leave the running checkers and membership intact.

The new regression covers SQL and SIGHUP, repeated loading and validation errors, membership changes in two groups, recovery with a valid config, changed group queries, group removal and re-addition, and normal shutdown.

Fixes #1611


test/functional/tests/group-reload-invalid/config.conf
```diff
@@ -0,0 +1,79 @@
+daemonize no
+workers 2
+# Reload parsing with ASAN needs more space than the default system stack.
```
krylosov-aa, line 3: The regression fixture uses a larger system coroutine stack to avoid an ASAN stack overflow. But we can remove this if #1611 is accepted.

PR #1613, opened 2026-09-20
Fix stack overflow during config reload

krylosov-aa:
SQL `RELOAD` and `SIGHUP` can crash Odyssey when the configuration contains nested includes. Each included file starts another parser call, which adds to stack usage. Reload runs on a small system coroutine stack, so it can overflow, especially in Debug and ASAN builds. The same configuration may load successfully at startup because startup uses the regular thread stack.

This change uses flex buffer stack to read included files within a single parser call. It preserves the include depth limit, error locations, and file boundaries: an unfinished directive or section in an included file cannot continue in its parent. The generated config parser and lexer also use `-O2` in Debug and sanitizer builds to reduce their stack usage.

This showed up on #1612


sources/cfg/scan.l
```diff
@@ -139,7 +140,9 @@ SIZE_SUFFIX  [kK][bB]?|[mM][bB]?|[gG][bB]?|[bB]
 
 
 \"([^\"]|\\\")*\" {
-					assert(yyleng >= 2);
+					if (yyleng < 2) {
+						return ERROR_TOKEN;
+					}
```
krylosov-aa, line 145: GCC 11 with LTO fails to link odyssey_test because it thinks yyleng - 2 can underflow and cause a malloc(SIZE_MAX) call. Flex guarantees at least two quotes, but GCC cannot see that, and the assert is disabled in Release builds.

The new test calls the config parser and exposes this warning.

PR #1615, opened 2026-09-21
fix: count actual number of hosts while iterating strtok

e-ivkov:
The count was derived from commas, but strtok_r collapses consecutive separators, so an empty or trailing-comma host left trailing entries uninitialised and od_storage_parse_endpoints passed indeterminate pointers to free(). Now reports the number of entries actually produced and also rejects an empty host.


sources/tests/odyssey/test_address.c
```diff
@@ -144,6 +144,17 @@ void odyssey_test_address_parse(void)
...
+
+	do_test("", NOT_OK_RESPONSE, 0);
+	do_test(",", NOT_OK_RESPONSE, 0);
+	do_test(",,", NOT_OK_RESPONSE, 0);
+
+	do_test("a,", OK_RESPONSE, 1, address("a", "", 0));
+	do_test(",a", OK_RESPONSE, 1, address("a", "", 0));
+	do_test("a,,b", OK_RESPONSE, 2, address("a", "", 0),
+		address("b", "", 0));
+	do_test(",a,,b,", OK_RESPONSE, 2, address("a", "", 0),
```
rkhapov, line 156: should this string be really successfully parsed?

e-ivkov: strtok_r already successfully parses it, as it collapses the separators. To prohibit this behavior we need to add additional checks. My PR's aim is to not pass incorrect pointers to free. If we want to change the strtok behavior or implement additional checks, I think we should do it in a separate PR

